PR #12, opened 2022-03-01
Throw error with mandatory options

SferaDev:
If the parameters are not set or the env variables are ``undefined``﻿, the  client gives a message that fetch expects and absolute url, instead adding a more meaningful error message.

Since it's a public facing API we should do this validations, in case the SDK is used from JavaScript instead of TS.

### DIFF
--- a/client/src/index.ts
+++ b/client/src/index.ts
@@ -305,9 +305,6 @@ export class RestRepository<T> extends Repository<T> {
 
   async request(method: string, path: string, body?: unknown) {
     const { databaseURL, apiKey } = this.client.options;
-    if (!databaseURL || !apiKey) {
-      throw new Error('Options databaseURL and apiKey are required');
-    }
 
     const resp: Response = await this.fetch(`${databaseURL}${path}`, {
       method,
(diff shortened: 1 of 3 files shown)


client/src/index.ts
```diff
@@ -304,13 +304,17 @@ export class RestRepository<T> extends Repository<T> {
   }
 
   async request(method: string, path: string, body?: unknown) {
-    const { databaseURL } = this.client.options;
+    const { databaseURL, apiKey } = this.client.options;
+    if (!databaseURL || !apiKey) {
+      throw new Error('Options databaseURL and apiKey are required');
+    }
```
SferaDev, line 310: Not sure if these conditions should be enforced on the constructor instead of the request. Is there any chance we would like to also offer a ``request(method: string, path: string, body?: unknown, options: {})`` that allows to overwrite one of these options?

gimenete: I'd prefer to enforce them as early as possible. For these ones in the constructor.

For the branch. We could validate in the constructor that the parameter is present. Then, when making the first request, validate that at least one strategy is returning a value.

In the future maybe we can expose a `request()` method, if our APIs grow fast and people want to use APIs not covered by the SDK. But, I don't see it happening soon, and I don't think we want to allow users to overwrite the XataClient options.

gimenete (commented):
Nice changes!

gimenete (approved):
LGTM 👌

Issue #11, opened 2022-02-28
Add ``local`` and ``git`` resolvers for ``branch`` option


gimenete:
The SDK resolves the branch the first time it needs it. It goes one by one, until it finds a string or a function that resolves in a stiring. Example:

```ts
import {local, git} from '@xata.io/client-utils/branch-resolution'

branch: [process.env.XATA_BRANCH, local, git, 'main']
```

In this case `local` and `git` are async function that return string, undefined or null. The resolved branch name will be stored in the XATA_BRANCH env variable. If it's not defined, it'll resolve the functions sequentially. If none of them return a defined string, the branch name used will be `main`.

SferaDev:
Closing as we've already implemented branch resolution in a different way

PR #13, opened 2022-03-01
Add branch strategy options

SferaDev:
Partially implements: https://github.com/xataio/client-ts/issues/11

- Remove ``branch`` from ``databaseUrl``
- Allow branch to be an array of values and/or be defined by a sync/async function

Future improvements: Add helper methods to load branch names from ``git`` or ``fs``

### DIFF
--- a/client/src/index.test.ts
+++ b/client/src/index.test.ts
@@ -1,24 +1,134 @@
-import { BaseClient, RestRepository, XataError, XataRecord } from './';
-
-const fetch = jest.fn();
-const client = new BaseClient(
-  {
-    fetch,
-    apiKey: '1234',
-    databaseURL: 'https://my-workspace-5df34do.staging.xatabase.co/db/xata',
-    branch: 'main'
-  },
-  {}
-);
+import { BaseClient, RestRepository, XataClientOptions, XataError, XataRecord } from './';
 
 interface User extends XataRecord {
   name: string;
 }
 
-const users = new RestRepository<User>(client, 'users');
+const buildClient = (options: Partial<XataClientOptions> = {}) => {
+  const {
+    apiKey = '1234',
+    databaseURL = 'https://my-workspace-5df34do.staging.xatabase.co/db/xata',
+    branch = 'main'
+  } = options;
+
+  const fetch = jest.fn();
+  const client = new BaseClient({ fetch, apiKey, databaseURL, branch }, {});
+  const users = new RestRepository<User>(client, 'users');
+
+  return { fetch, client, users };
+};
+
+describe('client', () => {
+  test('api key option is set', () => {
+    const { client } = buildClient({ apiKey: 'apiKey', databaseURL: 'url' });
+    expect(client.options.apiKey).toBe('apiKey');
+    expect(client.options.databaseURL).toBe('url');
+  });
+
+  test('throws if mandatory options are missing', () => {
+    // @ts-expect-error Options cannot be null in TypeScript
+    const { users } = buildClient({ apiKey: null, databaseURL: null });
+    expect(users.request('GET', '/foo')).rejects.toThrow('Options databaseURL and apiKey are required');
+  });
+
+  test('throws if branch cannot be resolved', () => {
+    const { users } = buildClient({ branch: null });
+
+    expect(users.request('GET', '/foo')).rejects.toThrow('Option branch is required');
+  });
+
+  test('provide branch as a string', async () => {
+    const { fetch, users } = buildClient({ branch: 'branch' });
+
+    fetch.mockReset().mockImplementation(() => {
+      return {
+        ok: true,
+        json: async () => ({})
+      };
+    });
+
+    await users.request('GET', '/foo');
+
+    expect(fetch).toHaveBeenCalledTimes(1);
+    expect(fetch.mock.calls[0]).toMatchInlineSnapshot(`
+      Array [
+        "https://my-workspace-5df34do.staging.xatabase.co/db/xata:branch/foo",
+        Object {
+          "body": undefined,
+          "headers": Object {
+            "Accept": "*/*",
+            "Authorization": "Bearer 1234",
+            "Content-Type": "application/json",
+          },
+          "method": "GET",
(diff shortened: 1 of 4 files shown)


client/src/index.test.ts
```diff
@@ -25,7 +26,7 @@ describe('request', () => {
       };
     });
 
-    users.request('GET', '/foo');
+    await users.request('GET', '/foo');
```
SferaDev, line 29: I'd like some kind of eslint rule to avoid this go unnoticed (ie: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md)

client/src/index.ts
```diff
@@ -470,3 +479,30 @@ export class XataError extends Error {
...
+    } else {
+      throw new Error(`Invalid branch strategy: ${strategy}`);
+    }
+  };
+
+  const strategy = strategies.find((option) => getBranch(option) !== undefined);
+  if (strategy) return await getBranch(strategy);
+
+  return undefined;
+};
```
SferaDev, line 508: Not 100% about code style preferences (type guards, typeof, helper methods and project structure, etc...), feel free to make comments about anything that should be different.

client/src/index.ts
```diff
@@ -470,3 +479,30 @@ export class XataError extends Error {
...
+      return await strategy();
+    } else if (isBranchStrategy(strategy)) {
+      return strategy;
+    } else {
+      throw new Error(`Invalid branch strategy: ${strategy}`);
+    }
+  };
+
+  const strategy = strategies.find((option) => getBranch(option) !== undefined);
+  if (strategy) return await getBranch(strategy);
```
SferaDev, line 505: I tried to make the branch strategy look-up lazy and stop on the first not-undefined value, however we are calling it twice now. Although the actions here should never be destructive, they could be slow. Not sure if it's worth to refactor it to a reduce that stops early but does not call it twice (less readable) or mutating a local variable in a loop (I personally try to avoid them as code style, but I'd like your feedback on how you normally approach these implementation details).

client/src/index.ts
```diff
@@ -470,3 +481,22 @@ export class XataError extends Error {
...
+  const strategies = Array.isArray(param) ? [...param] : [param];
+
+  const getBranch = async (strategy: BranchStrategy) => {
+    return isBranchStrategyBuilder(strategy) ? await strategy() : strategy;
+  };
+
+  for await (const strategy of strategies) {
+    const branch = await getBranch(strategy);
+    if (branch) return branch;
+  }
```
SferaDev, line 499: I made the early return to avoid heavy tasks to be run if a valid strategy was already found. Not 100% sure yet if you are for/against this kind of practices.

client/src/index.ts
```diff
@@ -470,3 +481,22 @@ export class XataError extends Error {
 }
 
 export type Links = Record<string, Array<string[]>>;
+
+const isBranchStrategyBuilder = (strategy: BranchStrategy): strategy is BranchStrategyBuilder => {
+  return typeof strategy === 'function';
+};
```
SferaDev, line 487: Would like input on how you normally use ``typeof`` and/or type guards. I tend to avoid ``typeof`` and only use them inside a type guard.

client/src/index.ts
```diff
@@ -304,12 +304,17 @@ export class RestRepository<T> extends Repository<T> {
...
+    const { databaseURL, apiKey, branch: branchStrategy } = this.client.options;
     if (!databaseURL || !apiKey) {
       throw new Error('Options databaseURL and apiKey are required');
     }
 
-    const resp: Response = await this.fetch(`${databaseURL}${path}`, {
+    const branch = await getBranchFromStrategy(branchStrategy);
+    if (!branch) {
+      throw new Error('Option branch is required');
+    }
```
SferaDev, line 315: This could be a cost consuming operation, should we memoize it?

client/src/index.test.ts
```diff
@@ -1,31 +1,142 @@
...
+    databaseURL = 'https://my-workspace-5df34do.staging.xatabase.co/db/xata',
+    branch = 'main'
+  } = options;
+
+  const fetch = jest.fn();
+  const client = new BaseClient({ fetch, apiKey, databaseURL, branch }, {});
+  const users = new RestRepository<User>(client, 'users');
+
+  return { fetch, client, users };
+};
```
SferaDev, line 19: Refactored the ``client``/``users`` a little so they're built with a helper instead of globals (allows for tests to customize them).

client/src/index.ts
```diff
@@ -397,9 +402,15 @@ export class RestRespositoryFactory implements RepositoryFactory {
   }
 }
 
+type BranchStrategyValue = string | undefined | null;
+type BranchStrategyBuilder = () => BranchStrategyValue | Promise<BranchStrategyValue>;
+type BranchStrategy = BranchStrategyValue | BranchStrategyBuilder;
+type BranchStrategyOption = BranchStrategy | BranchStrategy[];
```
SferaDev, line 408: Decided to be verbose here to make it more readable, but maybe I went too far.

client/src/index.ts
gimenete: I think we can just remember this value in a new field in `XataClient`. Then we'd do something like this:

```ts
this.branch = this.branch || await getBranchFromStrategy(branchStrategy);
```

client/src/index.ts
gimenete: I like early returns 😄

client/src/index.ts
gimenete: 👍 sounds good to me.

client/src/index.test.ts
gimenete: Agreed. This would be great. In the frontend code as well. I'll make a note and create an issue soon.

gimenete (commented):
This looks great!

I think we should implement the "memoization" before merging this PR. See my comment below.

gimenete (approved):
Nice work!! ✨

Issue #16, opened 2022-03-02
Add CI to the repository


gimenete:
- Use GitHub Actions to run tests in every push.
- Add secrets to run these tests against xata.io/sh.
- Prevent merging a PR if tests are not passing.

SferaDev:
Two topics I'd like to raise here:

#### Compiling each package

Right now the ``client-ts`` handles two npm packages (the client itself and the codegen). I could either manually target each package (maybe even with separate actions that check the changed files path to avoid compiling both packages if only one was touched) or add something more complex to handle a workspace (ie: turborepo, nx, ~lerna~, ...).

#### Publishing canary builds

I think something that adds value is having the PRs published on ``npm`` as canary builds (only if they're from a team member or a team member approves the deployment).

Also if we decide that we won't bump the versions every time a commit to main happens (and instead schedule releases with a set of features), when pushing to main a new beta could be published (that can be tested and used meanwhile the next release happens). This discussion is related to #17

gimenete:
I don't have experience with turborepo, nx, lerna, etc. If it's easy to use something like that for CI and publishing, great. If it's gonna take some time, I'd prefer to first implement just CI: GitHub Actions, running tests on every push. I think we should test both packages always, because if we've changed something in the codegen we could've broken the compatibility with the client and viceversa.

I like the idea of canary builds. Like I said, depending on how complex it is, I'd choose impelmenting only CI for now, or everything at the same time.

Issue #17, opened 2022-03-02
Implement a release strategy


gimenete:
So far I've been doing `npm publish` manually, but we should automate that.

fabien0102:
https://github.com/google-github-actions/release-please-action

For ref, this is what I'm usually using for this kind of problem

SferaDev:
@fabien0102 I very much enjoyed how the GitHub ``primer`` team does it.

In every PR you create a changeset file (markdown with frontmatter) where you set the description of the change for the release notes and the type to know how to bump the version number (ie: patch, minor, major). A bot helps you to create that file.

When the PRs are merged a bot is keeping track of them and opening another PR for ``Release Tracking``. In the moment you merge it, the changeset files are removed, the version is bumped following semver according to the changes included, the changelog file is updated, the version is pushed to ``npm`` and the release is tagged on GitHub.

What do you think of that approach?

Example: https://github.com/primer/react/pull/1894

TejasQ:
Given that #19 is merged and we have automatic canary publishes _and_ a manual workflow, is this closeable? Or is there more to be done here?

gimenete:
I think we can close it.

@SferaDev WDYT?

SferaDev:
I would like to have a more concise release strategy when actual users begin using the library.

For example, a bot that keeps track of the changes and decides which semver bump is needed and prepares the release notes.

Maybe we can keep this open as a low priority task but that needs to be implemented before public beta?

gimenete:
Sounds good 👍

gimenete:
I've never used semantic commits such as feat, chore, etc. but I can learn. No strong opinions.

Agreed that having the ability to PATCH a record is a main blocker. I think that was close to be done in the backend, right? And agreed that also links (and objects!) is the other high priority.

> Before that point we would need to begin documenting breaking changes and making sure we follow semver.

👍

PR #19, opened 2022-03-03
Add CI gh-action

SferaDev:
### Summary

Closes: #16 

Tasks:

- Add node version default (same as ``fronted-next``)
- Add codeowners and PR template
- Linter
- Tests
- Compile each module with lerna
- In PRs, publish canary build to NPM
- In main pushes, publish beta build to NPM
- Add a manual release action meanwhile we #17

#### Environment variables

We need to add the environment variables for integration tests in GitHub config. Not sure if we want to have a ``xata`` workspace shared with the team for that? (Useful for debugging)

#### PR checks

Also we need to configure that PRs cannot be merged if the actions fail. This can be done, after merging but we need to remember to do it.


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+*	@xataio/frontenders
```
gimenete, line 1: This is giving an error in the PR diff. I was thinking that maybe the team is secret but we have this in its configuration

image

🤔

gimenete (commented):
Looking good!

SferaDev:
Environment variables pending to be set:

- ``INTEGRATION_TEST_DATABASE_URL``
- ``INTEGRATION_TEST_API_KEY``
- ``INTEGRATION_TEST_BRANCH``
- ``NPM_TOKEN``
- ``GITHUB_TOKEN`` (Should be set automatically)

Haven't tried the comment step (unable locally). It's nothing fancy, if it works we keep it, if it doesn't I'll remove it.

.github/CODEOWNERS
TejasQ: Does the team have write access?

TejasQ (approved):
Cool, but why Lerna instead of something like Nx or TurboRepo?

SferaDev:
> Cool, but why Lerna instead of something like Nx or TurboRepo?

Fabien had experience with it and it was very simple to set up. We don't really have a complex build system or dependencies right now so most benefits of ``nx`` or ``turborepo`` are not needed (yet).

.github/CODEOWNERS
gimenete: That was it. Thanks!

gimenete:
> We need to add the environment variables for integration tests in GitHub config. Not sure if we want to have a xata workspace shared with the team for that? (Useful for debugging)

Yeah, I think we should do that. Go ahead and create a workspace and try to share with at least the people in the frontenders team.

gimenete:
@SferaDev I've configured the GitHub Actions secrets and I've created the database and its structure here https://app.xata.io/workspaces/sdk-integration-tests-19v8n2/dbs/test/main/tables/teams

SferaDev:
I re-ran an old job and pushed a new commit but it still fails on the integration test.

Notice I named the secrets slightly different than the actual env variables:

```
          XATA_DATABASE_URL: ${{ secrets.INTEGRATION_TEST_DATABASE_URL }}
          XATA_API_KEY: ${{ secrets.INTEGRATION_TEST_API_KEY }}
          XATA_BRANCH: ${{ secrets.INTEGRATION_TEST_BRANCH }}
```

gimenete:
@SferaDev I think it's because `test/integration.test.ts` is using `XATA_DATABASE_BRANCH` and the workflow file is using `XATA_BRANCH`.

For reference, these are the secrets we have 

image

TejasQ (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #21, opened 2022-03-03
Improve the output for code generation a bit

TejasQ:
This PR provides a DX with a bit more output for the user to keep them updated.

|Main|This PR|
|---|---|
|||

### DIFF
--- a/codegen/src/index.ts
+++ b/codegen/src/index.ts
@@ -1,15 +1,11 @@
 #!/usr/bin/env node
 import { program } from 'commander';
-import { join } from 'path';
+import { join, relative } from 'path';
 import ora from 'ora';
-import { access as lameCallbackBasedAccess } from 'fs';
-import { promisify } from 'util';
+import { access } from 'fs/promises';
 
 import { generate } from './codegen';
-import { relative } from 'path/posix';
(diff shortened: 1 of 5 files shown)


codegen/src/index.ts
```diff
@@ -1,8 +1,15 @@
 #!/usr/bin/env node
 import { program } from 'commander';
 import { join } from 'path';
+import ora from 'ora';
+import { access as lameCallbackBasedAccess } from 'fs';
+import { promisify } from 'util';
 
 import { generate } from './codegen';
+import { relative } from 'path/posix';
```
fabien0102, line 9: I see that we are using `join` from `path` instead of `path/posix`, let's merge the import statements!

codegen/src/index.ts
```diff
@@ -1,8 +1,15 @@
 #!/usr/bin/env node
 import { program } from 'commander';
 import { join } from 'path';
+import ora from 'ora';
+import { access as lameCallbackBasedAccess } from 'fs';
+import { promisify } from 'util';
```
fabien0102, line 6:
```suggestion
import { access } from 'fs/promises';
```

gimenete (approved):
LGTM 👌

PR #23, opened 2022-03-04
Seamlessly integrate with xataio/cli

TejasQ:
This PR updates the codegen to pull a Xata schema if one doesn't already exist since this is critical to using the SDK.

## Todo
- [x] Handle case where download from GitHub fails
- [x] Handle case where unzip/untar fails
- [x] Handle case where CLI executable is out of date, both on PATH and in tmp
- [x] Handle auth
- [x] Handle Xata Golang CLI auth failure
- [x] Test on Windows


codegen/src/useCli.ts
```diff
@@ -0,0 +1,34 @@
...
+    const cli = spawn(cliPath, ['init'], { stdio: ['inherit', 'pipe', 'pipe'] })
+      .on('close', (code) => {
+        if (code !== 0) {
+          return;
+        }
+        resolve();
+      })
+      .on('error', (e: unknown) => {
+        reject(e);
+      });
```
gimenete, line 19: nitpick:

```suggestion
      .on('error', reject);
```

codegen/src/useCli.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    cli.stdout.pipe(process.stdout);
+    cli.stdout.on('data', async (c) => {
+      const line = c.toString();
+      if (line !== 'Xata CLI is not configured, please run `xata auth login`\n') {
```
gimenete, line 24: This would break if the xata CLI changes the message. Suggestions:
- Use a regex to be more flexible and accept small changes in the message.
- Make the xata cli return a specific error code and use that in the `close` event.

codegen/src/getCli.ts
```diff
@@ -0,0 +1,70 @@
...
+    );
+
+  if (!fileUrl) {
+    failWithIncompatibleOs({ spinner });
+    return;
+  }
+
+  spinner.text = 'Downloading latest Xata CLI...';
+  const bundleName = fileUrl.split('/').slice(-1)[0];
+  const bundle = await fetch(fileUrl).then((f) => f.buffer());
```
gimenete, line 40: Non-blocking: I guess the xata cli is not going to be huge in size, but to save memory we could pipe directly the download to a file.

codegen/src/getCli.ts
```diff
@@ -0,0 +1,70 @@
...
+  );
+  spinner.start('Looking up latest Xata CLI...');
+  const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/latest')
+    .then((r) => r.json())
+    .then((d: GitHubResponse) =>
+      d.assets.map((a) => a.browser_download_url).find((a) => a.includes(getCliPlatformFromNodePlatform()))
+    );
+
+  if (!fileUrl) {
+    failWithIncompatibleOs({ spinner });
```
gimenete, line 34: We should print a different message when the list of download urls is empty. This can happen if a release is not generated correctly or their files are being uploaded and are not ready.

codegen/src/getCli.ts
TejasQ: Let's cross that bridge if/when we come to it. 🙌

gimenete (approved):

PR #24, opened 2022-03-04
Fix create object methods return type

SferaDev:
- Add link transform to ``createMany``
- Change ``create`` methods signature to return full objects


SferaDev:
@gimenete What do you think about changing the method signature to the ``id`` instead of the full object? If we return the full objects, we would need to ``get`` them from the API with an extra network call.

gimenete:
@SferaDev 🤔 I think the backend should return the full object on create. That way we'll get the object back, without an extra network call. I can see use cases where the backend will sanitize or create values that we don't have (e.g. created_at timestamps and such).

Do you want to start a discussion with the backend team?

SferaDev:
> @SferaDev 🤔 I think the backend should return the full object on create. That way we'll get the object back, without an extra network call. I can see use cases where the backend will sanitize or create values that we don't have (e.g. created_at timestamps and such).
> 
> Do you want to start a discussion with the backend team?

Sure, will open a discussion and ping them!

gimenete:
I'm leaning towards the idea of fetching the object doing another request for now. Looks like we'll have the option to get the full object back in the endpoint in the near future. So I prefer to fetch it and return it, rather than returning just the id.

The other option that also doesn't break the method signatures is merging the data in the SDK, instead of refetching. That's fine too, but I personally prefer the source of truth to be the backend.

SferaDev:
Okay, will do that and we can finally merge this.

gimenete (approved):
LGTM 👌

PR #29, opened 2022-03-15
Always fetch the CLI

TejasQ:
## Summary

I had a local `xata` that was outdated and it caused problems during codegen. Given that `codegen` is not something that is frequently run, let's always fetch the latest Xata CLI and use it for codegen.

What problems you had during codegen?

### Status Quo

Currently, the codegen CLI works like this on first run:

- Check for Xata schema. If found, generate. 
    - If not found, ask a user if they want to pull a schema via the Xata CLI.
        - If yes, check for a `xata` command in the user's current PATH
            - If found, execute `xata init` and follow the (potentially outdated) Xata CLI's `init` flow to pull down a local schema
            - If not found, ask if they want to download the Xata CLI
                - If yes, fetch the latest release from GitHub, untar into `/tmp` and execute it. They follow the latest Xata CLI's `init` flow to pull down a local schema, generate code, and exit
                - If no, exit
        - If no, exit
    - If found, generate code and exit.

On subsequent runs, where `xata/schema.json` is present, it generates code and exit, assuming the user will `xata pull` as needed.

### What this PR changes

This PR skips the entire "check if local Xata CLI exists" step on first run, always using the latest version of the Xata CLI. Subsequent runs of the codegen command are identical to main.

- Check for Xata schema. If found, generate. 
    - If not found, ask a user if they want to pull a schema via the Xata CLI.
        - Fetch the latest release from GitHub, untar into `/tmp` and execute it. They follow the latest Xata CLI's `init` flow to pull down a local schema, generate code, and exit
        - If no, exit
    - If found, generate code and exit

### Why I Feel It's Necessary

This PR only affects the first run, and has the following benefits:

1. It removes assumptions about the user's environment and treats it as hostile by default. This gives us the best chance of first run success by using the latest Xata CLI release and not running into edge cases where their system might cause bugs.
2. It does not augment any behavior with codegen _after_ the first run.

### Tradeoffs

- **Pro**: The user always uses the most up-to-date version of the Xata CLI to pull down a schema on their first codegen attempt, without the potential for version mismatches and other issues that might prove hard to support. I imagine we'd need to ask them to run `which xata` or similar to check the version, etc. in GitHub support issues, which might be more friction-forward than we want.
- **Con**: If GitHub is down or they're having network issues, they will not be able to download and use the Xata CLI. This con also exists without this feature, as one way or another, they will still need to get a `xata/schema.json` from somewhere.

### DIFF
--- a/codegen/src/getCli.ts
+++ b/codegen/src/getCli.ts
@@ -8,7 +8,6 @@ import { Ora } from 'ora';
 import { createReadStream, createWriteStream } from 'fs';
 
 import { cliPath } from './cliPath';
-import chalk from 'chalk';
 import { getCliPlatformFromNodePlatform } from './getCliPlatformFromNodePlatform';
 
 type GitHubResponse = {
@@ -16,13 +15,6 @@ type GitHubResponse = {
 };
 
 export const getCli = async ({ spinner }: { spinner: Ora }) => {
-  console.log();
-  spinner.info(
-    `You don't have the Xata CLI installed on your system, so we're downloading it. You could make this process faster by installing the CLI locally. More info: ${chalk.blueBright(
-      'https://docs.xata.io/cli/getting-started'
-    )}.
-`
-  );
   spinner.start('Looking up latest Xata CLI...');
   const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/latest')
     .then((r) => r.json())
--- a/codegen/src/getCli.ts
+++ b/codegen/src/getCli.ts
@@ -16,24 +16,19 @@ type GitHubResponse = {
 
 export const getCli = async ({ spinner }: { spinner: Ora }) => {
   spinner.start('Looking up latest Xata CLI...');
-  const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/latest')
-    .then((r) => r.json())
+  const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/skizzy')
+    .then(async (r) => {
+      if (!r.ok) {
+        throw errors.noCli;
+      }
+      return r.json();
+    })
     .then((d: GitHubResponse) =>
       d.assets.map((a) => a.browser_download_url).find((a) => a.includes(getCliPlatformFromNodePlatform()))
     );
 
   if (!fileUrl) {
-    spinner.fail(
-      `Could not find an appropriate version of the Xata CLI. This could be because:
-      
-1. No Xata CLI could be found for this platform (${process.platform}). 
-2. A release of the CLI was not correctly generated.
-3. A new release of the CLI is currently being rolled out.
-
-Please open an issue at https://github.com/xataio/cli and we'll address this as soon as we can. We apologize for the inconvenience.
-`
-    );
-    return;
+    throw errors.noCli;
   }
 
   spinner.text = 'Downloading latest Xata CLI...';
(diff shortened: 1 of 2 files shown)


codegen/src/getCli.ts
```diff
@@ -8,21 +8,13 @@ import { Ora } from 'ora';
...
 type GitHubResponse = {
   assets: { browser_download_url: string }[];
 };
 
 export const getCli = async ({ spinner }: { spinner: Ora }) => {
-  console.log();
-  spinner.info(
-    `You don't have the Xata CLI installed on your system, so we're downloading it. You could make this process faster by installing the CLI locally. More info: ${chalk.blueBright(
-      'https://docs.xata.io/cli/getting-started'
-    )}.
```
SferaDev, line 23: Can we still show a message if 1) CLI not found installing it or 2) CLI outdated, updating it?

gimenete:
What problems you had during codegen?

If I understand correctly with these changes the codegen is always going to download the CLI? I think that's a bit overkill. It can fail if the user doesn't have connectivity, or github is down, and it's going to slow down the process of code generation. If a user wants to test different things quickly, this would be a problem.

What if we check the current installed version with the latest version and we tell the user if it's outdated and ask them if they want to upgrade, like in other cli applications? Also, the result of querying which version is the last one should be cached for a day or so.

codegen/src/getCli.ts
TejasQ: Updated. 🙌

TejasQ:
Thanks a lot for your comments, everybody! I apologize for assuming the motivation, intent, and details of the proposed changes in this PR were clear. I've updated the description and I hope these things are a bit more clear now.

Now to respond to some of the comments:

## All

Thank you, all input is valid and helpful in creating a great experience for our users. 🙌 

## @SferaDev 

I've [responded](https://github.com/xataio/client-ts/pull/29#discussion_r827050218) specifically to the code comments inline.

## @gimenete 

> What problems you had during codegen?

The main problem was codegen failed and it was a little hard to understand why. After digging a bit, I identified the issue was indeed the way we infer the `xata` binary, which this PR aims to fix.

> If I understand correctly with these changes the codegen is always going to download the CLI? I think that's a bit overkill. It can fail if the user doesn't have connectivity, or github is down, and it's going to slow down the process of code generation. 

Thanks a lot for your thoughts. I've upgraded the description a bit with a bit more context. To clarify, with these changes the codegen is **only going to download the CLI on first run**, if the user has no `xata/schema.json`. I've highlighted more the tradeoffs if GitHub is down et al in the description.

> If a user wants to test different things quickly, this would be a problem.

I'd love to hear more about this and understand what a user would test differently quickly. Do you mean rapidly iterate on schema changes? In this case, we will not download the Xata CLI every time they run codegen. There is currently no flow defined for this as far as I can see, either on `main`, or in this PR, but I would love to discuss this further and create some type of `--watch` mode where we sync the local `xata/schema.json` with a remote. This seems like a great discussion for the future, and I don't think it falls under the scope of this PR.

> What if we check the current installed version with the latest version and we tell the user if it's outdated and ask them if they want to upgrade, like in other cli applications? Also, the result of querying which version is the last one should be cached for a day or so.

This is a great idea! If we do this in the Xata CLI, then we can close this PR. Actually, that's probably the simplest solution so we can do that as it would require no changes here. 👍 

Closing in favor of https://github.com/xataio/cli/issues/30.

## @tsg 

Thanks for the link! 🙌

TejasQ:
Opened ^ to preserve the useful feedback from @SferaDev from this PR.

PR #30, opened 2022-03-15
Fix local test flow and fetcher fallback

TejasQ:
## Summary

This PR fixes a number of issues. Each commit is easily cherry-pickable into more PRs if we want to break this up. The issues fixed are:

1. There is a certain rule we have disabled [on other projects](https://github.com/xataio/frontend-next/pull/422#discussion_r799035177) because it causes friction during development. Specifically, this rule is `@typescript-eslint/ban-ts-comment`. Sometimes `// @ts-ignore` is very much required and preferred over `// @ts-expect-error`, especially when in a test environment where we explicitly want to test against loosely typed cases. 

    With this rule enabled, we have to have an `// eslint-disable` comment just above `// @ts-ignore`, and as you can imagine, things get needlessly verbose so I propose disabling it in this PR for my test case. It was [disabled in frontend-next](https://github.com/xataio/frontend-next/pull/422#discussion_r799035177) for the same reason.

2. The integration test fails out of the box locally because the environment variables from `.envrc` are not correctly inferred by Node. The second commit fixes this.

3. The inference of `fetch` has caused [issues](https://xata-hq.slack.com/archives/C032EKCBUGM/p1646674428097719?thread_ts=1646668632.828269&cid=C032EKCBUGM) in the past, so I removed it.


gimenete:
> The integration test fails out of the box locally because the environment variables from .envrc are not correctly inferred by Node

Yeah, these variables are exported using [direnv](https://direnv.net/), which is the tool we use also in the `xata` repo, and we are using here. I don't have a strong opinion on using direnv or dotenv, but using direnv was simpler because it doesn't change anything in our code or in the repo.

Btw, it's a mistake we have `.envrc` tracked by git!

test/integration.test.ts
```diff
@@ -1,5 +1,10 @@
 import { contains, lt } from '../client/src';
 import { XataClient } from '../codegen/example/xata';
+import dotenv from 'dotenv';
+import { join } from 'path';
+
+// Get environment variables before reading them
+dotenv.config({ path: join(process.cwd(), '.envrc') });
```
gimenete, line 7: Does this work? Because an `.envrc` file doesn't have the same format than a `.env` file.

gimenete:
Btw, were you able to reproduce the issue with the SDK not using the `fetch` implementation you were passing to the constructor?

test/integration.test.ts
TejasQ: It does work, else I wouldn't have committed lol 🙌

TejasQ:
> Btw, were you able to reproduce the issue with the SDK not using the fetch implementation you were passing to the constructor?

Yes, multiple times, but I've since done a workaround and failed to capture the state of that specific repo.

client/src/index.ts
```diff
@@ -280,22 +282,14 @@ export class RestRepository<T> extends Repository<T> {
     super(null, table, {});
     this.client = client;
 
-    const { fetch } = client.options;
+    const doWeHaveFetch = typeof fetch !== 'undefined';
```
gimenete, line 285: What do you think if we check there's a `fetch` key in the options object and then do something if it's `undefined`

> const doWeHaveFetch = 'fetch' in client.options;
> if (typeof client.options.fetch === 'undefined') // the user wanted to pass something, but it's undefined. Wrong import/require() ?

gimenete:
> Yes, multiple times, but I've since done a workaround and failed to capture the state of that specific repo.

> I'll remove the fetch inference and just throw if no fetch with a link to the docs.

But I think that issue is not related to the fetch inference. Could be a problem somewhere else where we are using the global `fetch` even if there's a `this.fetch` defined inside `XataClient`? I'd like to find the root cause for that.

TejasQ:
> But I think that issue is not related to the fetch inference. Could be a problem somewhere else where we are using the global fetch even if there's a this.fetch defined inside XataClient? I'd like to find the root cause for that.

I'll commit next time I experience that and fork a repro repo. 👍

client/src/index.ts
TejasQ: Sure, why not? Something like e1d5b16f26d2dfce04d2920fa44b1933d1633508?

test/integration.test.ts
gimenete: I was just double checking. We can all commit things that don't work by mistake.

client/src/util/errors.ts
```diff
@@ -0,0 +1,12 @@
+export const errors = {
+  noFetchImplementation:
+    new Error(`The Xata client has no fetcher configured, and there is no global \`fetch\` function available for it to use. Please supply one in its constructor.
+    
+More in the docs: 
+` /** @todo add a link after docs exist */),
+  falsyFetchImplementation:
+    new Error(`The \`fetch\` option passed to the Xata client is resolving to a falsy value and may not be correctly imported.
```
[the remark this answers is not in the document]
TejasQ: Thanks a lot, this is done in 1987ee6ced81de7fc5aa5ff9587b1d06be6c2977.

gimenete (approved):

PR #31, opened 2022-03-15
Throw if Xata CLI not found

TejasQ:
## Summary

@SferaDev had great input in https://github.com/xataio/client-ts/pull/29#discussion_r826893541, so this PR is a subset of that one with this valuable change.

### DIFF
--- a/codegen/src/getCli.ts
+++ b/codegen/src/getCli.ts
@@ -25,7 +25,7 @@ export const getCli = async ({ spinner }: { spinner: Ora }) => {
   );
   spinner.start('Looking up latest Xata CLI...');
   const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/skizzy')
-    .then(async (r) => {
+    .then((r) => {
       if (!r.ok) {
         throw errors.noCli;
       }
--- a/codegen/src/getCli.ts
+++ b/codegen/src/getCli.ts
@@ -24,7 +24,7 @@ export const getCli = async ({ spinner }: { spinner: Ora }) => {
 `
   );
   spinner.start('Looking up latest Xata CLI...');
-  const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/skizzy')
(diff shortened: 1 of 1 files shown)


codegen/src/getCli.ts
```diff
@@ -24,24 +24,19 @@ export const getCli = async ({ spinner }: { spinner: Ora }) => {
 `
   );
   spinner.start('Looking up latest Xata CLI...');
-  const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/latest')
-    .then((r) => r.json())
+  const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/skizzy')
+    .then(async (r) => {
```
gimenete, line 28: Nitpick you can completely ignore. This function doesn't need to be async.

```suggestion
    .then((r) => {
```

codegen/src/getCli.ts
```diff
@@ -24,24 +24,19 @@ export const getCli = async ({ spinner }: { spinner: Ora }) => {
 `
   );
   spinner.start('Looking up latest Xata CLI...');
-  const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/latest')
-    .then((r) => r.json())
+  const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/skizzy')
```
gimenete, line 27: I guess you changed the URL to test the error scenario but forgot to put it back to the right URL.

```suggestion
  const fileUrl = await fetch('https://api.github.com/repos/xataio/cli/releases/latest')
```

gimenete (changes requested):
One blocking comment. The URL is not correct. Other than that LGTM.

codegen/src/getCli.ts
TejasQ: Hahahahhaha really good catch. I was testing the failure case. 😂  Thank you for this!

gimenete (approved):
:shipit:

PR #32, opened 2022-03-15
Use the xata command if exists locally

TejasQ:
## Summary

Fixing a bug where the codegen CLI always tries to use the Xata CLI from `tmpPath/xata`, even if the `xata` command is in a user's PATH and they've never downloaded the Xata CLI to `tmpPath`.


gimenete (approved):
LGTM 👌

PR #34, opened 2022-03-15
Unit test code generation

### LINKED ISSUE #33: Unit test code generation
As far as I can see, there are no tests for the codegen part, which makes it prone to regressions and similar issues. Let's add unit tests to this to ensure predictability. I'm happy to do this myself since we're all otherwise quite busy.

TejasQ:
# Why we need this

Considering future changes to codegen, this PR fixes #33 and adds unit tests to ensure we don't break it and it behaves as `expect`ed.


test/codegen.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+describe('codegen', () => {
+  it('should generate correct TypeScript', async () => {
+    const schemaFilePath = join(__dirname, 'mocks', 'schema.json');
+    const outputFilePath = 'hahaha';
+
+    const writeFile = jest.fn();
+    await generate({ schemaFilePath, outputFilePath, writeFile });
+
+    // Sorry about this.
```
gimenete, line 12: What if we use `toMatchSnapshot` instead of adding and using `toMatchWithoutWhitespace`?

We would remove all the expected outputs from this file and it'd be much cleaner.

TejasQ: Sure, great idea!

.gitignore
```diff
@@ -6,7 +6,6 @@ yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
 xata
-schema.json
```
gimenete, line 9: Question. Did you remove this for a reason?

gimenete (approved):
LGTM. Just one question.

.gitignore
TejasQ: Yes, because we need [this file](https://github.com/xataio/client-ts/pull/34/files#diff-ddbac6e328728ee2a7a8b36e7f2e73be149d74294c8925fdada37d5d24fd35b2).

Since we ignore the `xata` folder in the line directly above this one, we still meet the need of ignoring generated `schema.json`s while keeping intentional ones for tests and co.

PR #38, opened 2022-03-16
Fix #35: Use Xata folder instead of schema.json

### LINKED ISSUE #35: [Codegen] Shall we allow users to pass in a `xata` directory and work with this instead of schema.json?
Given that the Xata CLI creates a folder and puts multiple files in it (config.json, schema.json, and .gitignore), should we update the codegen CLI to receive an argument for the folder instead and then work with that?

I think this also might make sense since we’d like to use the config.json file according to #25.

TejasQ:
## Summary

This PR fixes #35 so we can use `config.json` as well in the near future.


README.md
```diff
@@ -20,12 +20,12 @@ Add an npm script to your package.json file to invoke `xata-codegen`. You can cu
...
 {
   "scripts": {
-    "xata-codegen": "xata-codegen -s xata/schema.json -o xata.ts"
+    "xata-codegen": "xata-codegen xata -o generated/client.ts"
   }
 }
 ```
 
-Run it (this is assuming your schema file is in `xata/schema.json` and it was created by the `xata` cli application)
+Run it (this is assuming your Xata folder is in `./xata` and it was created by the `xata` cli application)
```
gimenete, line 28: The `folder` term is more common on UI apps, whereas the `directory` term is more used in command line apps. WDYT about using `directory` instead of `folder`

gimenete: Yeah, but I mean, not only here but in all references to `folder`.

TejasQ: Sure, one sec...

TejasQ: Done in bite-sized pieces in https://github.com/xataio/client-ts/pull/38/commits/d242e2990c16bf66b2dc19198c814be94d11a21b and https://github.com/xataio/client-ts/pull/38/commits/fa80ed34370f83677a63e26784b5f6540ed1c7a9. 😄

gimenete (approved):
LGTM 👌

PR #44, opened 2022-03-16
Make databaseUrl optional in client after codegen

### LINKED ISSUE #25: Make `databaseURL` optional and inject it using the information in `xata/config.json` during code generation
Now that we have a `xata/config.json` file (we didn't have it a few months ago), we can read the workspace id and the database name during code generation and inject it in the code.

Users should still be able to overwrite its value for a few reasons:
- So the URL is configured with mechanisms like env variables. Useful for example if for security reasons somebody wants to run a codebase in a different workspace/database, or because it's a forked open source project and they don't want to change the generated code every time just to change the workspace/database, etc.
- For us, to test our generated code against localhost or staging.

TejasQ:
## Summary

This PR fixes #25 by making `databaseUrl` and inferring it from `config.json`. It also abstracts some of the file helpers to account for reading more than one file.


codegen/src/parseFile.ts
```diff
@@ -0,0 +1,39 @@
...
+import { xataDatabaseSchema, XataDatabaseSchema } from './schema';
+
+type Options = {
+  type: 'schema' | 'config';
+  input: string;
+};
+
+export function parseFile({ type, input }: Options & { type: 'config' }): XataConfigSchema;
+export function parseFile({ type, input }: Options & { type: 'schema' }): XataDatabaseSchema;
+export function parseFile({ type, input }: Options) {
```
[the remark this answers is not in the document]
TejasQ: Sure, done in https://github.com/xataio/client-ts/pull/44/commits/15c84855a6d48d42d98bcf1e32d3a3fc0bc37a41.

codegen/src/parseConfigFile.ts
```diff
@@ -0,0 +1,11 @@
+import { xataConfigSchema } from './config';
+import { handleParsingError } from './handleParsingError';
+
+export const parseConfigFile = (input: string) => {
+  try {
+    return xataConfigSchema.parse(JSON.parse(input));
+  } catch (err) {
+    handleParsingError(err);
+    throw err; // ^ runs process.exit(1) if successful. If not, let's throw the error.
```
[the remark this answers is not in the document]
TejasQ: We can, but then this function would return `XataConfigSchema | undefined` and we'd have to optionally chain and account for potential `undefined` returns in `codegen.ts`. Happy to do this if you like, let me know. 🙌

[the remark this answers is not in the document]
TejasQ: Great question. I'll clarify my comment to answer this directly in the code.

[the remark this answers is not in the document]
TejasQ: Awesome. It's in https://github.com/xataio/client-ts/pull/44/commits/ee4047cdfbeb4209e5ec6641d16f3bfb418aaa1f. Let me know if you'd like to go the optional chain route and I'll make that happen too.

codegen/src/codegen.ts
```diff
@@ -134,15 +113,18 @@ export async function generate({
...
     ${tables.map((table) => generateTableType(table)).join('\n')}
 
     const links = ${JSON.stringify(links)};
 
     export class XataClient extends BaseClient<{
       ${tables.map((table) => `"${table.name}": Repository<${getTypeName(table.name)}>;`).join('\n')}
     }> {
-      constructor(options: XataClientOptions) {
-        super(options, links);
+      constructor(options: ${config.dbName ? `PartialBy<XataClientOptions, 'databaseURL'>` : 'XataClientOptions'}) {
```
gimenete, line 126: I thought `config.dbName` would be always defined. Is that not the case?

TejasQ: Great question. I'm thinking ahead a bit here for cases where the Xata CLI (outside of our control) changes and `dbName` may not be the same key or similar. Happy to remove this if you think it's simpler. Let me know.

gimenete: 🤔  I'm not sure if I follow. As far as I see in the code, if the key changes, the zod validation would fail, right? (I'm not super familiar with zod, but looks like you'd have to use `z.string().optional()` if you want to cover that case).

See https://github.com/xataio/client-ts/pull/44/files#diff-74e5a79804f41cdb60666c9eab8560b0f6ab27bf8f1936b5058e3a90f36e03bbR5

Anyway I don't think we should cover that case. It is very unlikely that the format will change in that way, and in that case the validation during parsing should fail.

In order to prevent backwards-incompatible changes in configuration files, I think the best practice is using a version number inside the file itself (such as `lockfileVersion` in `package-lock.json`).

TejasQ: Cool, done in https://github.com/xataio/client-ts/pull/44/commits/5bef124beaaa4d3d4cd0c76b7f43a957841dd556.

test/__snapshots__/codegen.test.ts.snap
```diff
@@ -79,6 +79,9 @@ exports[`codegen should generate correct TypeScript 1`] = `
   XataRecord,
 } from \\"@xata.io/client\\";
 
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
```
gimenete, line 83: I've testing locally and I think this adds a bit of complexity to the user:

Code in main:

image

Code in this PR:

image

[We talked](https://github.com/xataio/client-ts/issues/25#issuecomment-1069061003) about just making the `databaseURL` field optional:


```diff
  export type XataClientOptions = {
    fetch?: unknown;
-   databaseURL: string;
+   databaseURL?: string;
    branch: BranchStrategyOption;
    apiKey: string;
    repositoryFactory?: RepositoryFactory;
  };
```

I think that's simpler and there are no side effects. The constructor of `BaseClient` already checks if databaseURL is defined. So that's the only change required and we can get rid of the PartialBy.

I'm happy to make this change directly in this PR if you want.

TejasQ: I respectfully disagree because:

1. When comparing the developer experience of main vs. this PR, both work quite well with autocompletion, with the `PartialBy` case making it quite explicit what's missing (`apiKey` and `branch`). Here's a side-by-side comparison, where both cases guide the user by telling them what exactly is missing, and providing autocompletion options:

|Main|This PR|
|---|---|
|||

2. `XataClientOptions` in main is pretty much as opaque as its `PartialBy` counterpart in this PR,

3. We did talk about `databaseURL` being optional, it is now.

For these reasons, I believe there is no extra complexity to the user because they are clearly guided by autocompletion and type hints as in the [GIF above](https://user-images.githubusercontent.com/9947422/158775471-4d60b7c3-d84c-491f-8b55-a180e5f2fb22.gif). Still, if you insist, feel free to make any changes this PR that you believe is best. 🙌

gimenete (approved):

PR #46, opened 2022-03-16
Add internal openapi schemas and fetchers

SferaDev:
## Related issues or discussions

OpenAPI usage proposal: https://github.com/xataio/frontend-next/discussions/586#discussioncomment-2364140

Code structure issue: https://github.com/xataio/client-ts/issues/50

## Motivation

- Use ``openapi`` spec to generate API operations, fetchers, schemas and types.
- Allow performing schema-less operations from the SDK
- Structure the code into smaller files

### Split ``databaseURL`` into ``workspace`` and ``database``

```diff
export type XataClientOptions = {
- fetch?: unknown;
+ fetch?: FetchImpl;
- databaseURL?: string;
+ workspace?: string;
+ database?: string;
  branch: BranchStrategyOption;
  apiKey: string;
  repositoryFactory?: RepositoryFactory;
};
```

Similar to ``frontend-next`` add two new global environment variables (to me this should be also options of the client).

```diff
+export MAIN_API_BASE_URL=https://staging.xatabase.co
+export WORKSPACES_API_BASE_URL=https://{workspaceId}.staging.xatabase.co
```

### Add new API namespace of the client

```
const xata = new BaseClient(options);
xata.api.records.insertRecord(workspace, database, branch, this.table, record);
xata.api.workspace.createWorkspace({ name: "foo", slug: "foo" });
```

### For the schema operations internally use the API client

We still offer, and promote in the docs, the schema-generated client. Instead of having hardcoded fetch calls, it now internally uses the ``xata.api`` automatically typed.


client/src/api/xatabaseFetcher.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (e instanceof Error) {
+      const error: SimpleError = {
+        message: e.message,
+        status: 500
+      };
+      throw error;
+    } else if (
+      typeof e === 'object' &&
+      typeof (e as SimpleError).message === 'string' &&
+      typeof (e as SimpleError).status === 'number'
```
fabien0102, line 102: You need to remove this check, the backend don't send `status` anymore ;)

client/src/api/xatabaseFetcher.ts
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable @typescript-eslint/no-throw-literal */
+/* eslint-disable @typescript-eslint/ban-types */
+import type { SimpleError } from './xatabaseResponses';
+
+const resolveUrl = (
+  url: string,
+  _queryParams: Record<string, unknown> = {},
```
fabien0102, line 7: [Just pointing the obvious to not forget] We need a `qs` implem here!

client/src/api/xatabaseFetcher.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (response.ok) {
+      return jsonResponse;
+    }
+
+    throw jsonResponse;
+  } catch (e) {
+    if (e instanceof Error) {
+      const error: SimpleError = {
+        message: e.message,
+        status: 500
```
fabien0102, line 98: Careful, `SimpleError` don't have `status` anymore.

Ref:
https://github.com/xataio/frontend-next/pull/656

client/src/api/xatabaseFetcher.ts
SferaDev: Yeah, I thought of that. We should build our own query string serializer as we aim to be 0-dependency. I would've imported ``qs`` library otherwise.

client/src/api/xatabaseFetcher.ts
SferaDev: Done!

client/src/api/xatabaseFetcher.ts
fabien0102: https://nodejs.org/dist/latest-v16.x/docs/api/url.html#class-urlsearchparams

Also available in nodejs API

SferaDev:
As discussed, closing in favor of smaller PRs that adapt to the new decision of structure

PR #52, opened 2022-03-17
Derive codegen language from output file path

### LINKED ISSUE #51: Codegen language selection is not working properly
From the README:

```
xata-codegen generate [path to xata directory] -o generated/client.ts
```

If you do that in `main`, the result is a `generated/client.ts.ts` file. Note the duplicated extension

If you do

```
xata-codegen generate [path to xata directory] -o generated/client.js
```

The result is a `generated/client.js.ts` extension.

Previously the language was deducted from the extension. But now there's an optional --lang parameter that if it's not used, the extension the `.ts` extension is always added regardless of the extension the user provides. I haven't checked what happens if you use the param and also put an extension in the `-o` option.

My suggestion is to remove the --lang option and deduct the language from the extension.

cc @TejasQ 

It's easy to reproduce the probem by runnning `npm run build-example`

image

TejasQ:
## Summary

This PR fixes #51 by deriving the language from the output file path and removing the now unnecessary language parameter.


gimenete (commented):
> now unnecessary language parameter.

I don't recall when that was necessary, but that doesn't matter now.

Changes LGTM. Before approving, can you run `npm run build-example` and commit the changes?

TejasQ:
> I don't recall when that was necessary

I assume you thought it was necessary when you wrote it in this early commit you authored: https://github.com/xataio/client-ts/blob/d2d0c2d1f93f029672acf9e0534a84220aa99bb8/codegen/src/codegen.ts#L95

To be clear, this parameter is removed now because `output` is sufficient.

> Changes LGTM. Before approving, can you run npm run build-example and commit the changes?

Running and adding to pre-commit...

gimenete:
> I assume you thought it was necessary when you wrote it in this early commit you authored: 

I didn't introduce the command line parameter for the `language`. The `language` parameter in the function was not obtained from the CLI arguments. It was calculated from the output parameter.

https://github.com/xataio/client-ts/blob/d2d0c2d1f93f029672acf9e0534a84220aa99bb8/codegen/src/index.ts#L22-L23

TejasQ:
If `generate` itself gets `output`, why is it necessary that it needs `language`?

gimenete:
> If generate itself gets output, why is it necessary that it needs language?

You can argument that the language calculation could have been done inside the `generate` function, but what we are discussing is that the `language` parameter as a CLI parameter was not necessary.

I found the commit that introduced the `language` argument in the CLI: https://github.com/xataio/client-ts/commit/925e139d4de402796a4a2993469cb427bd35a7c8#diff-84c035ea9096c494edcf4dcd1190fa67880a30d995e1622b34799e247062c11f

TejasQ:
> but what we are discussing is that the `language` parameter as a CLI parameter was not necessary.

Haha, sorry. My bad. I wasn't clear again. I meant the unnecessary language parameter to the `generate` function is now unnecessary, since I changed its signature in this PR. 🙌 

To be fair, it looks equally unnecessary even previously.

gimenete:
Well, I've been pretty clear that I was talking about "CLI arguments" in my comments. And the important part of this PR is deleting this

https://github.com/xataio/client-ts/pull/52/files#diff-84c035ea9096c494edcf4dcd1190fa67880a30d995e1622b34799e247062c11fL36-L40

Anyway, I'll review again the PR.

codegen/src/getLanguageFromExtension.ts
```diff
@@ -0,0 +1,14 @@
+import type { Language } from './codegen';
+import { errors } from './errors';
+
+export const getLanguageFromExtension = (extension?: 'ts' | 'js') => {
```
gimenete, line 4: This function is missing the return type. Otherwise we are not using the `Language` type here in this file, even though it is imported.

```suggestion
export const getLanguageFromExtension = (extension?: 'ts' | 'js'): Language => {
```

gimenete: Also, the `Language` type can be simplified now to:

> export type Language = 'typescript' | 'javascript';

codegen/src/generateWithOutput.ts
```diff
@@ -1,24 +1,19 @@
...
 }: {
   spinner: Ora;
   xataDirectory: string;
   out: string;
-  lang: Language;
 }) => {
   spinner.text = 'Found schema, generating...';
 
-  await generate({ xataDirectory, outputFilePath: out, language: lang });
+  await generate({ xataDirectory, outputFilePath: out });
```
gimenete, line 16: Nitpick. This function could just have two arguments: xataDirectory and outpuFilePath. Not really necessary to have an object as arguments if only has two fields and both are required and always passed in the same order.

gimenete: You can absolutely ignore this nitpick.

package.json
```diff
@@ -43,6 +43,9 @@
     "typescript": "^4.5.5"
   },
   "lint-staged": {
+    "codegen/src/*.{js,ts,tsx}": [
+      "npm run build-example"
+    ],
```
gimenete, line 48: Nice! 👍

codegen/src/generateWithOutput.ts
TejasQ: I don't really have a strong opinion, but with object syntax, we get some benefits:

1. Autocompletion. If we call `generate()`, TS will tell us we need an object, if we run `generate({})`, TS will tell us `xataDirectory` and `outputFilePath` are missing.
2. We can be more liberal with order of options.
3. It's more explicit in that `xataDirectory` is always any given identifier (like `xataDirectory: anyVariable`).

We lose some of these as function arguments:

1. Order of arguments works, and breaks when we change the signature of this function. This is not a huge deal because we have TypeScript so it's still safe.
2. It can be a bit ambiguous to call `generate(arg1, arg2)` without the explicit key/value syntax.

Not a strong opinion, but let me know if you'd like me to change it and I will. 🙌

codegen/src/getLanguageFromExtension.ts
TejasQ: Thanks, done in https://github.com/xataio/client-ts/pull/52/commits/e5f4de371cc4a75cfc78bb3a074c6ec7c1785136.

TejasQ:
Something is very broken with lint-staged's commits (they're empty): https://github.com/xataio/client-ts/pull/52/commits/64b92971dc0688d3bb3bdcb940d16dde3662f63a

Investigating. Please check back later.

codegen/src/generateWithOutput.ts
gimenete: > Autocompletion

vscode will tell you the name of the arguments and their types. So, there's no difference here. You have to type more if the arguments are encapsulated in an object.

> We can be more liberal with order of options.

Sure, but in this case we have always two arguments and both are required.

> Not a strong opinion, but let me know if you'd like me to change it and I will. 🙌

No need to change it. But I think it's something we should follow in general. If the function receives very few arguments and all/most of them are optional, we should use "regular" arguments.

gimenete (approved):

gimenete:
> A review on the whole PR instead of some commits with potential errors will be much more efficient. 🙌

I review the whole PR, but if you commit changes, I need to review the changes in those new commits.

TejasQ:
> I review the whole PR, but if you commit changes, I need to review the changes in those new commits.

Why? Some commits have errors, and the whole PR always reflects the latest state. 🤔

gimenete:
If you leave a comment like this

image

You can expect me to click there and see the changes.

PR #54, opened 2022-03-18
Bundle size checker

SferaDev:
## Summary

- Add bundle size check
- Move exclude to a new tsconfig
- Remove files from bundle published on npm

## Notes

GitHub action will fail here as the HEAD branch does not have ``size-limit`` installed. See https://github.com/xataio/client-ts/pull/46#issuecomment-1072099116 for an example.

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -7,7 +7,7 @@ jobs:
     env:
       CI_JOB_NUMBER: 1
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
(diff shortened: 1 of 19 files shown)


.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,13 @@
+name: Bundle size
+on: [pull_request]
+jobs:
+  size:
+    name: Size limit
+    runs-on: ubuntu-latest
+    env:
+      CI_JOB_NUMBER: 1
+    steps:
+      - uses: actions/checkout@v1
```
gimenete, line 10:
```suggestion
      - uses: actions/checkout@v3
```

PR #55, opened 2022-03-19
Improve README

gimenete:
- We don't require() well-known libraries anymore.
- Remove the `databaseURL` in the example, but explain that it can be specified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ import { XataClient } from './xata';
 const client = new XataClient({
   branch: 'branchname',
   apiKey: 'xau_1234abcdef',
-  fetch: fetchImplememntation // Required if your runtime doesn't provide a global `fetch` function.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,7 +37,6 @@ In a TypeScript file start using the generated code:
 import { XataClient } from './xata';
 
 const client = new XataClient({
-  databaseURL: 'https://myworkspace-123abc.xata.sh/db/databasename',
   branch: 'branchname',
   apiKey: 'xau_1234abcdef',
   fetch: fetchImplememntation // Required if your runtime doesn't provide a global `fetch` function.
```
SferaDev, line 42:
```suggestion
  fetch: fetchImplementation // Required if your runtime doesn't provide a global `fetch` function.
```

SferaDev, line 42:
```suggestion
  fetch: fetchImplementation // Required if your runtime doesn't provide a global `fetch` function.
```

SferaDev (approved):

PR #56, opened 2022-03-20
Add Xata API Client

SferaDev:
Add Xata API Client constructor

```ts
const client = new XataApiClient({
  fetchImpl: fetch,
  apiKey: process.env.XATA_API_KEY,
  api: 'staging'
});

await client.workspace.getWorkspace("my-workspace")
```

Also re-export all the fetchers as functions so that they can be manually used in case the app wants to tree-shake the api (only using a small subset of the API).


client/src/api/index.ts
```diff
@@ -0,0 +1,8 @@
+import { operationsByTag } from './components';
+import type * as Responses from './responses';
+import type * as Schemas from './schemas';
+
+export * from './client';
+export * from './components';
+export { operationsByTag as Operations };
+export type { Responses, Schemas };
```
SferaDev, line 8: Official exports for apps:

1. ``XataApiClient`` constructor based client
2. All fetchers exposed as tree-shakeable functions
3. ``Operations`` dictionary by tags of the fetchers
4. ``Responses``and ``Schemas`` types

We might not want to promote all of them in the docs, maybe only the constructor based implementation, but I like to provide as much tools any experienced developer might want to use.

client/src/api/providers.ts
```diff
@@ -0,0 +1,34 @@
+import { errors } from '../util/errors';
+
+type AliasedProviders = 'production' | 'staging';
+type ProviderBuilder = { main: string; workspaces: string };
+export type ApiProviders = AliasedProviders | ProviderBuilder;
```
SferaDev, line 5: User can provide either their own provider builder (two different urls) or use ``production`` (default).

For development (we) will use ``staging``.

client/src/api/fetcher.ts
```diff
@@ -0,0 +1,120 @@
...
+  body?: TBody;
+  headers?: THeaders;
+  queryParams?: TQueryParams;
+  pathParams?: TPathParams;
+};
+
+const fallbackError: SimpleError = { message: 'Network response was not ok' };
+
+function baseURLForWorkspace(workspacesApiUrl: string, workspace: string) {
+  if (!workspacesApiUrl.includes('{workspaceId}')) return workspacesApiUrl;
```
SferaDev, line 38: I tried to take into account the scenario where they have CNAME links of their workspace url, but in the docs we will need to say that ``{workspaceId}`` is a restricted keyword that injects the workspace id.

client/src/api/fetcher.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  // Node.js on localhost won't resolve localhost subdomains unless mapped in /etc/hosts
+  // So, instead, we use localhost without subdomains, but will add a Host header
+  if (typeof window === 'undefined' && workspacesApiUrl.includes('localhost:')) {
+    return workspacesApiUrl.replace('{workspaceId}.', '');
+  }
```
SferaDev, line 44: Not 100% sure about this, haven't tried localhost backend yet.

client/src/api/fetcher.ts
```diff
@@ -0,0 +1,120 @@
...
+  const response = await fetchImpl(`${baseURL}${resolveUrl(url, queryParams, pathParams)}`, {
+    method: method.toUpperCase(),
+    body: body ? JSON.stringify(body) : undefined,
+    headers: {
+      'Content-Type': 'application/json',
+      ...headers,
+      Authorization: `Bearer ${apiKey}`,
+      // The host header is needed by Node.js on localhost.
+      // It is ignored by fetch() in the frontend
+      ...(pathParams?.workspace ? { Host: hostHeaderForWorkspace(workspacesApiUrl, pathParams.workspace) } : {})
```
SferaDev, line 83: Not 100% sure about this, haven't tried localhost backend yet.

client/src/api/fetcher.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+  return workspacesApiUrl.replace('{workspaceId}', workspace);
+}
+
+function hostHeaderForWorkspace(workspacesApiUrl: string, workspace: string) {
+  const url = baseURLForWorkspace(workspacesApiUrl, workspace);
+  if (!url) return;
+  const [, hostname] = url.split('://');
+  return hostname;
+}
```
SferaDev, line 53: Not 100% sure about this, haven't tried localhost backend yet.

client/src/api/fetcher.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  try {
+    const jsonResponse = await response.json();
+
+    if (response.ok) {
+      return jsonResponse;
+    }
+
+    if (jsonResponse.success) {
```
SferaDev, line 99: Not sure about this ``.success`` property, is it being still used?

openapi-codegen.config.ts
```diff
@@ -0,0 +1,83 @@
...
+                    schema: { type: 'string' }
+                  }
+                ]
+              }
+            ]
+          : [key, value]
+      )
+    )
+  };
+};
```
SferaDev, line 83: To be removed when the PR on ``openapi-codegen`` is merged, didn't want to point to ``file:``.

client/src/api/client.ts
```diff
@@ -0,0 +1,540 @@
...
+import type * as Responses from './responses';
+import type * as Schemas from './schemas';
+
+export interface XataApiClientOptions {
+  fetchImpl: FetchImpl;
+  apiKey: string;
+  api?: ApiProviders;
+}
+
+export class XataApiClient {
```
SferaDev, line 15: Didn't like ``XataRestClient`` and proposing ``XataApiClient``, no strong opinion. Open for discussion!

client/src/util/errors.ts
```diff
@@ -1,4 +1,6 @@
 export const errors = {
+  noFetchImplementation: 'No fetch implementation provided',
+  invalidApiProvider: 'Invalid API provider',
   falsyFetchImplementation: `The \`fetch\` option passed to the Xata client is resolving to a falsy value and may not be correctly imported.
```
SferaDev, line 4: Not sure about having an ``errors`` dictionary. Code looks cleaner but makes debugging harder. I saw a comment on a previous PR, not sure what was the outcome.

openapi-codegen.config.ts
```diff
@@ -0,0 +1,83 @@
...
+export default defineConfig({
+  xatabase: {
+    from: {
+      source: 'file',
+      relativePath: '../openapi/bundled/openapi.yaml'
+    },
+    outputDir: 'client/src/api',
+    to: async (context) => {
+      // TODO: Fix me, allow no filenamePrefix
+      const filenamePrefix = ' ';
```
SferaDev, line 13: I will open tomorrow a PR for this nasty workaround 🥲

client/src/api/fetcher.ts
```diff
@@ -0,0 +1,120 @@
+/* eslint-disable @typescript-eslint/no-throw-literal */
+/* eslint-disable @typescript-eslint/ban-types */
+import type { SimpleError } from './responses';
+
+const resolveUrl = (
+  url: string,
+  _queryParams: Record<string, unknown> = {},
```
SferaDev, line 7: Yes, I forgot to add the query string builder...

client/src/api/fetcher.ts
```diff
@@ -0,0 +1,120 @@
...
+  pathParams: Record<string, string> = {}
+) => {
+  return url.replace(/\{\w*\}/g, (key) => pathParams[key.slice(1, -1)]);
+};
+
+// Typed only the subset of the spec we actually use (to be able to build a simple mock)
+export type FetchImpl = (
```
SferaDev, line 14: This should be probably moved elsewhere, potentially used in other parts of the SDK

client/src/api/fetcher.ts
```diff
@@ -0,0 +1,120 @@
...
+    const jsonResponse = await response.json();
+
+    if (response.ok) {
+      return jsonResponse;
+    }
+
+    if (jsonResponse.success) {
+      throw withStatus(jsonResponse.data, response.status);
+    } else if (jsonResponse.message) {
+      throw withStatus({ message: jsonResponse.message }, response.status);
```
SferaDev, line 102: Instead of throwing on error I would prefer a union that can be discriminated with data and error objects inside, what do you think provides better DX?

client/src/api/client.ts
```diff
@@ -0,0 +1,540 @@
+import { errors } from '../util/errors';
+import type * as Types from './components';
+import { operationsByTag } from './components';
+import type { FetcherExtraProps, FetchImpl } from './fetcher';
+import { ApiProviders, getApiUrl } from './providers';
+import type * as Responses from './responses';
+import type * as Schemas from './schemas';
+
+export interface XataApiClientOptions {
+  fetchImpl: FetchImpl;
```
gimenete, line 10: Suggestion. We call this `fetch` in the other client. I think it's a bit nicer than `fetchImpl`.

client/src/api/client.ts
```diff
@@ -0,0 +1,540 @@
...
+  apiKey: string;
+  api?: ApiProviders;
```
gimenete, line 12: I think we could better use a `hostname`. People are not going to use it now, but may be confusing if they see `production` somewhere in our code and they can think this is about their production environment or something like that, or if they think `api` has something to do with API versioning.

And potentially in the future if we provide a local development server or if we provide on-prem xata instances, it makes more sense to have the hostname fully customizable. WDYT?

I know we have two different hosts in production (api.xata.io and xata.sh) but that's the exception. So I'd only add one hostname parameter, and if not specified, use the specific default one for each case (api.xata.io or xata.sh).

client/src/api/client.ts
gimenete: No strong opinion either. `XataApiClient` is nicer. I suggested `XataRestClient` because maybe in the future we have other APIs (GraphQL?, Twitch?).

client/src/api/client.ts
```diff
@@ -0,0 +1,540 @@
...
+  public get records() {
+    return new RecordsApi(this.extraProps);
+  }
+}
+
+class UserApi {
+  constructor(private extraProps: FetcherExtraProps) {}
+
+  public getUser(): Promise<Schemas.UserWithID> {
+    return operationsByTag.users.getUser({ ...this.extraProps });
```
gimenete, line 63: This is not auto-generated, right? How hard would it be to generate these in an automated way?

client/src/api/fetcher.ts
gimenete: I'm wondering if can use maybe the types that the TypeScript language defines in the "dom" lib:

https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L18664

gimenete (commented):
A quick review for now. I'll do another deeper review later.

client/src/api/client.ts
SferaDev: I thought of generating this, but I liked that we get to manually decide which changes we want to introduce to the public facing SDK API. ``tsc`` will type check that the wrapper matches the types from the ``auto-generated`` functions.

client/src/api/client.ts
SferaDev: Makes sense

client/src/api/client.ts
SferaDev: Sure thing

client/src/api/fetcher.ts
SferaDev: This enforces our users to add ``DOM`` to their ``tsconfig`` which I wanted to avoid. Also providing a subset of the spec we actually use to be able to build simple mocks.

client/src/api/fetcher.ts
gimenete: 🤔  I don't think that's in the spec. Not sure what's that coming from.

client/src/api/fetcher.ts
gimenete: No strong opinions. If you want leave a message for the future.

client/src/util/errors.ts
gimenete: No strong opinions. I prefer to have the message where it is used, because normally you have different errors for each case and that way you can quickly find where a message is being used.

client/src/api/client.ts
```diff
@@ -0,0 +1,555 @@
...
+
+    this.extraProps = {
+      apiUrl: getHostUrl(provider, 'main'),
+      workspacesApiUrl: getHostUrl(provider, 'workspaces'),
+      fetchImpl,
+      apiKey: options.apiKey
+    };
+  }
+
+  public get user() {
```
gimenete, line 34: What do you think about using plural in these getters? users, workspaces, databases?

client/src/api/client.ts
```diff
@@ -0,0 +1,555 @@
...
+}
+
+export class XataApiClient {
+  private extraProps: FetcherExtraProps;
+
+  constructor(options: XataApiClientOptions) {
+    const fetchImpl = typeof fetch !== 'undefined' ? fetch : options.fetchImpl;
+    if (!fetchImpl) {
+      throw new Error(errors.noFetchImplementation);
+    }
```
SferaDev, line 22: TODO: Review falsy fetch implementation error

SferaDev, line 22: TODO: Review falsy fetch implementation error

client/src/util/errors.ts
SferaDev: Since we are not consistent, removing them and inlining inside the error constructors

client/src/api/fetcher.ts
SferaDev: No need, it will be a breaking change and if we ever do it we will have to think about how to handle safe operations

client/src/api/fetcher.ts
SferaDev: Removed!

SferaDev:
Updated this PR with integration tests creating a specific ``database`` and removing the problem with concurrency.

gimenete (approved):
There's a lot to review here 😅 I've done a few passes and I don't have any suggestion or requsted change. Great work!

PR #57, opened 2022-03-21
Allow selecting/sorting columns

SferaDev:
Internal types are still a work in progress that needs to be cleansed. I will probably split a bit the code into smaller files and utility types (and document how/why inference works).

The following now just works and is type-safe:

image

I still need to review ``links`` in general (they won't probably work yet)

### DIFF
--- a/client/src/index.test.ts
+++ b/client/src/index.test.ts
@@ -290,7 +290,7 @@ describe('query', () => {
     test('simple query', async () => {
       const { users } = buildClient();
 
-      const expected = { method: 'POST', path: '/tables/users/query', body: {} };
+      const expected = { method: 'POST', path: '/tables/users/query', body: { columns: ['*'] } };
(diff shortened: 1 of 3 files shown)


SferaDev:
@gimenete Code is not ready to be merged, but I'd like a first review before cleaning-up the types

SferaDev:
> Nice! 💯 
> 
> I guess it's not possible to make the `select()` variadic, and for that reason we need an array, right?
> 
> Other than that, this looks great!

It's not possible to make a difference between select(undefined) and select([]) and that's why I used an array. Also it's consistent with the option in getOne/getMany.

gimenete:
Cool, makes sense 👍

SferaDev:
I was going to improve the typings and organize files but it would be a big diff. Splitting into smaller PRs, will merge this as it was reviewed.

Issue #59, opened 2022-03-22
Client SDK not working with Cloudflare Worker


ncheikh:
This line is not working for the following Cloudflare Worker config:

https://github.com/xataio/client-ts/blob/7fe10d6b3d936f9129408267e6cc6b9d4ea7551c/client/src/index.ts#L407

I was able to work around this by changing `this.fetch` to `fetch` at the above line.  I attempted:

1. To manipulate the constructor in RestRepository to assign cloudflare workers fetch - Did not work
2. Printing and comparing `fetch` and `this.fetch` and they report as the same.
3. Adjusting tsconfig to ES6 instead of esnext to align with client and validate this wasn't the issue - Did not work

I am really not sure why this isn't working, the only thing I added beyond the template was itty-router and faunadb (replacing with xata).  This was a problem with both local `wrangler dev` and deployed code.

Cloudflare Worker:
 - Template: https://github.com/cloudflare/worker-typescript-template

nvmrc
```
lts/gallium
```
 
tsconfig:
```
{
  "compilerOptions": {
    "outDir": "./dist",
    "module": "commonjs",
    "target": "esnext",
    "lib": ["esnext"],
    "alwaysStrict": true,
    "strict": true,
    "preserveConstEnums": true,
    "moduleResolution": "node",
    "sourceMap": true,
    "esModuleInterop": true,
    "types": [
      "@cloudflare/workers-types",
      "@types/jest",
      "@types/service-worker-mock"
    ]
  },
  "include": ["src"],
  "exclude": ["node_modules", "dist", "test"]
}
```

package.json
```
{
  "name": "tasker-backend-cf",
  "version": "0.1.0",
  "main": "dist/worker.js",
  "scripts": {
    "build": "webpack",
    "format": "prettier --write  '*.{json,js}' 'src/**/*.{js,ts}' 'test/**/*.{js,ts}'",
    "lint": "eslint --max-warnings=0 src && prettier --check '*.{json,js}' 'src/**/*.{js,ts}' 'test/**/*.{js,ts}'",
    "test": "jest --config jestconfig.json --verbose",
    "xata-codegen": "xata pull && xata-codegen generate -o src/generated/XataClient"
  },
  "author": "Nabil Cheikh",
  "eslintConfig": {
    "root": true,
    "extends": [
      "typescript",
      "prettier"
    ]
  },
  "devDependencies": {
    "@cloudflare/workers-types": "^3.0.0",
    "@types/jest": "^26.0.23",
[...]

PR #61, opened 2022-03-22
Fix fetch implementation with CF workers

SferaDev:
Closes: #59 

In #58 we plan to use the API methods internally and stop using ``fetch`` directly (which also solves the `Illegal Invocation` issue). This is only a fix so we can bump a ``patch`` version early and solve the compatibility with CF.

### DIFF
--- a/client/src/index.ts
+++ b/client/src/index.ts
@@ -378,7 +378,7 @@ export abstract class Repository<T> extends Query<T, Selectable<T>> {
 
 export class RestRepository<T> extends Repository<T> {
   client: BaseClient<any>;
-  fetch: any;
+  fetchImpl: any;
 
   constructor(client: BaseClient<any>, table: string) {
     super(null, table, {});
@@ -386,10 +386,10 @@ export class RestRepository<T> extends Repository<T> {
 
     const fetchImpl = typeof fetch !== 'undefined' ? fetch : this.client.options.fetch;
     if (!fetchImpl) throw new Error('No fetch implementation provided');
-    this.fetch = fetchImpl;
+    this.fetchImpl = fetchImpl;
 
     Object.defineProperty(this, 'client', { enumerable: false });
-    Object.defineProperty(this, 'fetch', { enumerable: false });
+    Object.defineProperty(this, 'fetchImpl', { enumerable: false });
     Object.defineProperty(this, 'hostname', { enumerable: false });
   }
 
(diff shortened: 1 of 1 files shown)


client/src/index.ts
```diff
@@ -403,8 +396,9 @@ export class RestRepository<T> extends Repository<T> {
   async request<T>(method: string, path: string, body?: unknown): Promise<T | undefined> {
     const { databaseURL, apiKey } = this.client.options;
     const branch = await this.client.getBranch();
+    const fetchImpl = this.fetch;
```
SferaDev, line 399: This is actually the fix 🤯

client/src/index.ts
```diff
@@ -384,16 +384,9 @@ export class RestRepository<T> extends Repository<T> {
...
-
-    if (doWeHaveFetch) {
-      this.fetch = fetch;
-    } else if (isInjectedFetchProblematic) {
-      throw new Error(errors.falsyFetchImplementation);
-    } else {
-      this.fetch = this.client.options.fetch;
-    }
+    const fetchImpl = typeof fetch !== 'undefined' ? fetch : this.client.options.fetch;
+    if (!fetchImpl) throw new Error('No fetch implementation provided');
```
gimenete, line 388: I see some value in this check:

```ts
    if (!fetchImpl && 'fetch' in this.client.options) throw new Error('Invalid fetch implementation provided');
```

That was the idea with the "falsy" error message.

In any case we should remove the `errors.falsyFetchImplementation` in the errors file if we don't use it anymore.

client/src/index.ts
gimenete: Wow! Great find!

Yeah, no need to bind IMO.

gimenete (approved):
Thanks for the fix!

ncheikh:
Thanks for the fix! Tested with this branch and my CF workers now behave :slightly_smiling_face:

client/src/index.ts
ncheikh: @SferaDev Could you please explain why this works vs just using this.fetch in the call?

client/src/index.ts
SferaDev: Since it's not related to the fix on CF, let's move the discussion to #56

client/src/index.ts
SferaDev: @ncheikh It's not a simple explanation 😢 

If you're curious I strongly suggest you to read [YDKJS 1st edition](https://github.com/getify/You-Dont-Know-JS/blob/1st-ed/this%20&%20object%20prototypes/README.md#you-dont-know-js-this--object-prototypes). 

What is weird is that it does work on node.js but not on V8 (CF Workers JS runtime), I would have to dig deeper to know the exact cause.

ncheikh: Thank you @SferaDev, I will take a look!  I was a bit baffled why it wasn't working and tried several things but it wasn't obvious to me why it wasn't working for CF as it was working in other environments/engines. I wanted to open an MR for the fix but couldn't come up with a solution and was hopeful y'all would be able to sort it out and add clarity.

I very much appreciate the quick turn around and taking the time to answer my question!

PR #62, opened 2022-03-23
Organize files

SferaDev:
This PR is the first step towards organising the files of the SDK a little bit.

We had three scenarios:

- Schema generated client: ``class XataClient extends XataSchemaClient<{ ... }>;``
- Schema-less client: ``const client = new XataSchemaClient()``
- API client: ``const api = new XataApiClient()``

**Proposed file structure:**

- api/
    - index.ts [Public exports for the ``XataApiClient``]
    - [file].ts
- schema/
    - index.ts [Public exports for the ``XataSchemaClient``]
    - [file].ts
- index.ts [Public exports of the library, reexporting ``api`` and ``schema``]


gimenete (approved):
LGTM 👌

Thank you for all the effort you are putting in the SDK!

PR #63, opened 2022-03-23
Make class properties private

SferaDev:
Reduce the number of properties that appear when using intellisense on a ``Repository``

### DIFF
--- a/client/src/index.ts
+++ b/client/src/index.ts
@@ -164,7 +164,7 @@ export class RestRepository<T extends XataRecord> extends Repository<T> {
   ): Promise<
     Page<T, typeof options['columns'] extends SelectableColumn<T>[] ? Select<T, typeof options['columns'][number]> : R>
   > {
-    const data = query.getData();
+    const data = query.getQueryOptions();
 
     const body = {
       filter: Object.values(data.filter).some(Boolean) ? data.filter : undefined,
--- a/client/src/index.ts
+++ b/client/src/index.ts
@@ -15,10 +15,6 @@ export interface XataRecord {
 }
 
 export abstract class Repository<T extends XataRecord> extends Query<T> {
-  select<K extends SelectableColumn<T>>(columns: K[]) {
-    return new Query<T, Select<T, K>>(this.repository, this.table, { columns });
-  }
-
   abstract create(object: Selectable<T>): Promise<T>;
 
   abstract createMany(objects: Selectable<T>[]): Promise<T[]>;
@@ -29,8 +25,7 @@ export abstract class Repository<T extends XataRecord> extends Query<T> {
 
   abstract delete(id: string): void;
 
-  // Used by the Query object internally
-  abstract _runQuery<R extends XataRecord, Options extends QueryOptions<T>>(
+  abstract query<R extends XataRecord, Options extends QueryOptions<T>>(
     query: Query<T, R>,
     options: Options
   ): Promise<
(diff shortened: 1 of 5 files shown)


client/src/index.ts
```diff
@@ -15,10 +15,6 @@ export interface XataRecord {
 }
 
 export abstract class Repository<T extends XataRecord> extends Query<T> {
-  select<K extends SelectableColumn<T>>(columns: K[]) {
-    return new Query<T, Select<T, K>>(this.repository, this.table, { columns });
-  }
```
SferaDev, line 20: ``Select`` is actually a method from ``Query`` which ``Repository`` extends. Moving to its correct place

client/src/index.ts
```diff
@@ -29,8 +25,7 @@ export abstract class Repository<T extends XataRecord> extends Query<T> {
 
   abstract delete(id: string): void;
 
-  // Used by the Query object internally
-  abstract _runQuery<R extends XataRecord, Options extends QueryOptions<T>>(
+  abstract query<R extends XataRecord, Options extends QueryOptions<T>>(
```
SferaDev, line 28: Since the parameters are public, expose ``query`` method instead of making it internal or protected. It's a ``READ`` operation in CRUD.

client/src/schema/query.ts
```diff
@@ -10,43 +11,121 @@ export type QueryOptions<T> = {
...
-
-    // For some reason Object.assign(this, parent) didn't work in this case
-    // so doing all this manually:
-    this.$any = parent?.$any;
-    this.$all = parent?.$all;
-    this.$not = parent?.$not;
-    this.$none = parent?.$none;
-    this.$sort = parent?.$sort;
-
-    Object.assign(this, data);
```
SferaDev, line 47: Hiding the internal ``filters`` into a private object instead of public ``class`` properties

client/src/schema/query.ts
```diff
@@ -58,101 +137,62 @@ export class Query<T extends XataRecord, R extends XataRecord = T> implements Ba
     Object.defineProperty(this, 'repository', { enumerable: false });
   }
 
+  getData(): QueryTableData {
+    return this.#data;
+  }
```
SferaDev, line 142: Add a public method to retrieve them from the ``Repository``

client/src/schema/query.ts
```diff
@@ -10,43 +11,121 @@ export type QueryOptions<T> = {
...
+} & {
+  [key: string]: FilterColumn;
+};
+
+export type FilterList = FilterExpression | FilterExpression[];
+export type FilterColumn = FilterColumnIncludes | FilterPredicate | FilterList;
+
+/**
+ * @maxProperties 1
+ * @minProperties 1
```
gimenete, line 30: Interesing! Didn't know this existed. Is this used by TypeScript and/or vscode?

client/src/schema/query.ts
gimenete: `data`/`getData` could confuse our users thinking that it's the way to execute the query.

What about something like "query options/parameters"

client/src/schema/query.ts
```diff
@@ -192,28 +232,20 @@ export class Query<T extends XataRecord, R extends XataRecord = T> implements Ba
...
     return this.getPaginated({ page: { size, offset } });
   }
 
   async lastPage(size?: number, offset?: number): Promise<Page<T, R>> {
     return this.getPaginated({ page: { size, offset, before: 'end' } });
   }
-
-  hasNextPage(): boolean {
-    return this.meta.page.more;
-  }
```
gimenete, line 218: Question. I see this method is being removed and `previousPage` as well. Can you explain why? If it's been moved I don't see where it is now.

gimenete (commented):
Looking great! I have just one question and one suggestion.

client/src/schema/query.ts
SferaDev: This is generated from the openapi spec, unfortunately it's not used in TS/VSCode.

I've done things like these with helpers before in TS (maybe as an improvement for the future):

```ts
export type NonEmptyArray<T> = T[] & { 0: T };

export type RequireAtLeastOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<keyof T, Keys>> &
    {
        [K in Keys]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<Keys, K>>>;
    }[Keys];
```

client/src/schema/query.ts
SferaDev: Done!

client/src/schema/query.ts
SferaDev: Yeah, those methods on a repository/query level did not make much sense. But its true that we were partially removing an easy way to do the following:

```ts
function *foo(page: Paginable) {
    yield page.records;
    if (page.hasNextPage()) return foo(page.nextPage());
}

foo(xata.db.users)
```

Added it back!

gimenete (approved):
Nice work!! ✨

Issue #26, opened 2022-03-11
Add support for PATCH operations


SferaDev:
Related to: https://github.com/xataio/xata/pull/583

PR #66, opened 2022-03-24
Implement PATCH operations in Update

SferaDev:
Closes: #26


client/src/schema/selection.ts
```diff
@@ -19,7 +20,8 @@ type OmitMethods<T> = {
   [key in keyof T as T[key] extends Function ? never : key]: T[key];
 };
 
-export type Selectable<T> = Omit<OmitQueries<OmitMethods<T>>, 'id' | 'xata'>;
+type InternalProperties = 'xata' | 'read' | 'update' | 'delete';
```
gimenete, line 23: This list may change. We know `xata` is always going to be there, but we may add more functions in the future. I had a type to omit functions, what do you think about using that?

gimenete (approved):
Looks good! Just one non-blocking comment.

client/src/schema/selection.ts
SferaDev: I have a better solution in mind:

```suggestion
type InternalProperties = keyof Omit<XataRecord, 'id'>;
```

PR #67, opened 2022-03-25
Fix connection with localhost

SferaDev:
Fix host header with localhost connection

### DIFF
--- a/client/src/api/fetcher.ts
+++ b/client/src/api/fetcher.ts
@@ -81,6 +81,8 @@ export async function fetch<
   const baseUrl = buildBaseUrl({ path, workspacesApiUrl, pathParams, apiUrl });
   const fullUrl = resolveUrl(baseUrl, queryParams, pathParams);
 
+  // Node.js on localhost won't resolve localhost subdomains unless mapped in /etc/hosts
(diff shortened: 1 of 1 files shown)


client/src/api/fetcher.ts
```diff
@@ -84,18 +79,18 @@ export async function fetch<
   workspacesApiUrl
 }: FetcherOptions<TBody, THeaders, TQueryParams, TPathParams> & FetcherExtraProps): Promise<TData> {
   const baseUrl = buildBaseUrl({ path, workspacesApiUrl, pathParams, apiUrl });
-  const url = resolveUrl(baseUrl, queryParams, pathParams);
+  const fullUrl = resolveUrl(baseUrl, queryParams, pathParams);
+
+  const url = fullUrl.includes('localhost') ? fullUrl.replace(/^[^.]+\./, 'http://') : fullUrl;
```
SferaDev, line 84:
```suggestion
  // Node.js on localhost won't resolve localhost subdomains unless mapped in /etc/hosts
  // So, instead, we use localhost without subdomains, but will add a Host header
  const url = fullUrl.includes('localhost') ? fullUrl.replace(/^[^.]+\./, 'http://') : fullUrl;
```

gimenete (approved):

PR #69, opened 2022-03-25
Add changesets for release tracking

SferaDev:
Closes: #17 

If we agree on it, we should install https://github.com/apps/changeset-bot

## Why ``changesets`` instead of ``release-please``

Forcing semantic commits on PR title and squash strategy might be enough for us, but we need to check it always for ourselves before merging a PR. With ``changesets`` we have a bot during PR that prepares for the "Release tracking" and it makes it easier. If we accept contributions from outsiders it will be much easier and we reduce human error.

## How it works

You can either add a changeset from the CLI by ``npx changeset`` or do it online with the help of the bot.

image
image
image
image

After a PR is merged but yet not released, the bot updates another PR with the ``Release tracking``. Once that ``Release tracking`` PR is merged it automatically publishes the package and creates a changelog. Do we want to customize the changelog?


SferaDev:
@fabien0102 Would love to hear your opinion on this 😄

.github/workflows/release.yml
```diff
@@ -41,15 +39,21 @@ jobs:
...
           XATA_DATABASE_URL: ${{ secrets.INTEGRATION_TEST_DATABASE_URL }}
           XATA_API_KEY: ${{ secrets.INTEGRATION_TEST_API_KEY }}
           XATA_DATABASE_BRANCH: ${{ secrets.INTEGRATION_TEST_BRANCH }}
 
-      - name: Bump version
-        run: npx lerna version ${{ github.event.inputs.bump }} --yes
-
-      - name: Publish build
-        run: npx lerna publish from-package --no-verify-access --yes
+      - name: Create Release Pull Request or Publish to npm
```
gimenete, line 52: Maybe we can be more specific here? We are not creating a PR, right?

gimenete (approved):
Looks good! Just a small suggestion

.github/workflows/release.yml
SferaDev: It actually does one or the other depending on the branch.

If it's being called from feature branches, it creates/updates the release tracking PR.

If it's being called from the main branch, it publishes to npm the new release and cleans up the old changesets files.

gimenete: Makes sense 👍 !

SferaDev:
Let's try this 🤞🏻

PR #73, opened 2022-03-28
Move final files to a subfolder

SferaDev:
We have been updating the internal structure of the folders in the SDK, this is the last part of such refactor.


SferaDev:
>15.09 KB (+27.35% 🔺)

Not sure if we should have:

```
import { X } from "@xata.io/client/api"
import { X } from "@xata.io/client/schema"
```

Issue #6, opened 2022-02-08
`filter` types are not accurate


fabien0102:
The `delete`, `read` and `update` action should not be valid as `filter` field.

![image](https://user-images.githubusercontent.com/1761469/153004438-aa140c42-8dc9-4d8a-82f8-e82e05352771.png)

PR #74, opened 2022-03-30
Improve internal types

SferaDev:
- Filter values should not show internal methods. Closes: https://github.com/xataio/client-ts/issues/6
- Make codegen fields ``nullable``
- Improve types for basic ``deno`` support


README.md
```diff
@@ -135,3 +135,20 @@ await user.delete();
 // Using an object's id
 await xata.db.users.delete('rec_1234abcdef');
 ```
+
+## Deno support
+
+Right now we are still not publishing the client on deno.land or have support for deno in the codegen.
+
+However you can already use it with your preferred node CDN with the following import in the auto-generated ``:
```
gimenete, line 143: Is something missing between these backticks?

gimenete (approved):
LGTM. Just a minor comment.

README.md
SferaDev, line 143:
```suggestion
However you can already use it with your preferred node CDN with the following import in the auto-generated ``xata.ts`` file:
```

SferaDev: Actually yes 🤯

README.md
```diff
@@ -135,3 +135,20 @@ await user.delete();
 // Using an object's id
 await xata.db.users.delete('rec_1234abcdef');
 ```
+
+## Deno support
```
gimenete, line 139:
```suggestion
## Deno support
```

There's a weird whitespace that prevents this line to be rendered as a heading. See https://github.com/xataio/client-ts/blob/f1e660da2d79ab2314266f0fb2e18055e944bbe2/README.md

README.md
```diff
@@ -135,3 +135,20 @@ await user.delete();
...
+
+Right now we are still not publishing the client on deno.land or have support for deno in the codegen.
+
+However you can already use it with your preferred node CDN with the following import in the auto-generated ``xata.ts`` file:
```
gimenete, line 143:
```suggestion
However you can already use it with your preferred node CDN with the following import in the auto-generated `xata.ts` file:
```

Maybe you put two double ticks on purpose? But I don't think it's necessary.

PR #75, opened 2022-04-01
Add documentation comments

gimenete:
This PR basically adds documentation comments. Useful for generating static pages with the API reference and also for contextual documentation in IDEs such as vscode.


client/src/schema/operators.ts
```diff
@@ -2,23 +2,75 @@ import { Constraint } from './filters';
...
 export const includes = (value: string): Constraint<string> => ({ $includes: value });
+/**
+ * Operator to restrict results to only arrays that include a value matching the given substring.
+ */
 export const includesSubstring = (value: string): Constraint<string> => ({ $includesSubstring: value });
+/**
+ * Operator to restrict results to only arrays that include a value matching the given pattern.
+ */
 export const includesPattern = (value: string): Constraint<string> => ({ $includesPattern: value });
 export const includesAll = (value: string): Constraint<string> => ({ $includesAll: value });
```
SferaDev, line 76: Missing comment here

client/src/schema/operators.ts
```diff
@@ -2,23 +2,75 @@ import { Constraint } from './filters';
...
+/**
+ * Operator to restrict results to only values that are not equal to the given value.
+ */
 export const isNot = <T>(value: T): Constraint<T> => ({ $isNot: value });
+/**
+ * Operator to restrict results to only values that contain the given value.
+ */
 export const contains = <T>(value: T): Constraint<T> => ({ $contains: value });
 
 // TODO: these can only be applied to columns of type "multiple"
```
SferaDev, line 63: All these filters should need to be reviewed, opened https://github.com/xataio/client-ts/issues/76.

Also we are missing some like ``includesNone``, but I prefer looking at these after this PR

client/src/schema/pagination.ts
```diff
@@ -26,23 +36,51 @@ export class Page<T extends XataRecord, R extends XataRecord> implements Paginab
     this.records = records;
   }
 
+  /**
+   * Retrieves the next page of results.
+   * @param size Maximum number of results to be retrieved.
+   * @param offset Number of results to skip when retrieving the results.
```
SferaDev, line 42: Applies to all methods, should we add the type to the param comment? (Usage outside TS)

SferaDev (approved):

client/src/schema/operators.ts
gimenete: I know. Because I wasn't sure how it works 😅

SferaDev: Yeah, it should be an array not a string the param hehe, it's fine as is

client/src/schema/pagination.ts
gimenete: Good suggestion. Not sure about that. I've just created an issue to decide that https://github.com/xataio/client-ts/issues/77

client/src/schema/operators.ts
gimenete: Since we are going to review operators in the issue you [opened](https://github.com/xataio/client-ts/issues/76) I'm going to leave this as it is for now. Good catch anyway!

PR #79, opened 2022-04-04
Update dependencies

SferaDev:
- Update changeset files (one change per file)
- Add action from ``frontend-next`` to update regularly dependencies
- Run manually the update


SferaDev:
We need to discuss, our stance on ESM modules. Some of the dependencies of the codegen are now ESM-only. Doesn't matter much for our client though.

SferaDev:
> @SferaDev I see you are downgrading those dependencies that are now ESM only. What's the technical reason for that?

We need to make some changes to support ESM only libraries, and I wanted us to discuss it first before changing anything (see previous message)

gimenete:
@SferaDev yeah, I read your previous message. But I still don't know what are the technical implications of using ESM libraries vs not using them 😄

gimenete (approved):
LGTM 👌

PR #80, opened 2022-04-04
Add a stale comment on build PR action

SferaDev:
When pushing code to an existing PR it's good to see that the package published is not up-to-date with the latest changes (while the github action is still compiling). Similar to what Vercel does with preview PRs.


gimenete (approved):
Nice!

PR #82, opened 2022-04-07
Difference between insert and create

SferaDev:
As discussed in Madrid station, let's differentiate between insert and create.


gimenete (commented):
LGTM. A few comments before approving:

- Could you add integration an integration test for the `insert` method?
- We should add the `insert` method to the `Repository` abstract class.
- What do you think about changing `upsert` to `updateOrInsert`?

gimenete (approved):
🚀

PR #87, opened 2022-04-08
Fix running codegen as an ES Module

SferaDev:
When did JS world became so hard to work with 😭

### DIFF
--- a/codegen/scripts/add-version-file.js
+++ b/codegen/scripts/add-version-file.js
@@ -1,0 +1,14 @@
+import fs from 'fs';
+import { fileURLToPath } from 'url';
+import path from 'path';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const pkgPath = path.join(__dirname, '../package.json');
+const pkgContents = fs.readFileSync(pkgPath, 'utf8');
+const pkg = JSON.parse(pkgContents);
+
+const contents = `export const CODEGEN_VERSION = '${pkg.version}';`;
+const versionPath = path.join(__dirname, '../src/version.ts');
+fs.writeFileSync(versionPath, contents);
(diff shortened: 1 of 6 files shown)


codegen/scripts/add-version-file.js
```diff
@@ -0,0 +1,14 @@
+import fs from 'fs';
+import { fileURLToPath } from 'url';
+import path from 'path';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
SferaDev, line 6: In ES Modules ``__dirname`` is not a global anymore, there might be better approaches. So far this made sense for those comming from CJS.

codegen/scripts/add-version-file.js
```diff
@@ -0,0 +1,14 @@
...
+
+const pkgPath = path.join(__dirname, '../package.json');
+const pkgContents = fs.readFileSync(pkgPath, 'utf8');
+const pkg = JSON.parse(pkgContents);
+
+const contents = `export const CODEGEN_VERSION = '${pkg.version}';`;
```
SferaDev, line 12: We have to inject the version in compile time because we don't want to import the ``package.json`` into the bundle. And no, we are unable to require it dynamically because it's no longer possible under ES Modules.

codegen/scripts/inject-hashbang.js
```diff
@@ -0,0 +1,16 @@
...
+import path from 'path';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const pathName = path.join(__dirname, '../dist/index.js');
+const original = fs.readFileSync(pathName, 'utf8');
+
+const contents = `#!/bin/sh
+':' //; exec /usr/bin/env node --es-module-specifier-resolution=node "$0" "$@"
```
SferaDev, line 12: We need this to tell node that we don't have ``.js`` endings in the import tags

fabien0102 (approved):
👌

Issue #64, opened 2022-03-23
Consider generating TypeScript and then transpiling it to JavaScript when generating JS code


gimenete:
Right now the codegenerator can generate JavaScript, but maybe it's a better option to generate TypeScript and then transpile it. That could even be better to ensure that tools such as vscode find the types correctly.

My only concern is how well the generated code looks, because the generated code is something people may look at. Anyway maybe we can use prettier after the transpilation?

cc @fabien0102

SferaDev:
Done in https://github.com/xataio/client-ts/pull/88 😄

PR #88, opened 2022-04-08
Allow codegen to be invoked as a library

SferaDev:
Closes: #64 
Required for: SDK Shell

- Move command line to ``cli``
- Refactor ``codegen`` so it doesn't directly read/write file system
- Export useful methods as a library
- Remove the ``js`` generation and transpile it from TypeScript code


codegen/src/codegen.ts
```diff
@@ -125,42 +108,33 @@ export async function generate({ outputFilePath: output, xataDirectory, writeFil
...
+};
 
-  const pretty = prettier.format(code, { parser: 'babel' });
-  await writeFile(fullOutputPath, pretty);
+function transpile(code: string, language: Language, javascriptTarget: JavascriptTarget = 'ES2020') {
+  switch (language) {
+    case 'typescript':
+      return code;
+    case 'javascript':
+      return ts.transpile(code, { target: ts.ScriptTarget[javascriptTarget] });
```
fabien0102, line 138: 👌

fabien0102 (approved):
Looks good

codegen/src/codegen.ts
SferaDev: I knew you'd like this 🤖

PR #96, opened 2022-04-12
Implement getAll method

SferaDev:
Closes: #91


client/src/schema/query.ts
```diff
@@ -201,24 +199,37 @@ export class Query<T extends XataRecord, R extends XataRecord = T> implements Pa
...
+   * Warning: If there are a large number of results, this method can have performance implications.
+   * @param options Additional options to be used when performing the query.
+   * @returns An array of records from the database.
+   */
+  async getAll<Options extends QueryOptions<T>>(
+    options: Omit<Options, 'page'> = {} as Options
+  ): Promise<GetWithColumnOptions<T, R, typeof options>[]> {
+    const results = [];
+
+    for await (const page of this.getIterator(100, options)) {
```
gimenete, line 218: Could we extract this number to a constant?
Question: is 100 the limit of the API?

gimenete (approved):

client/src/schema/query.ts
SferaDev: Actually no, moving discussion to #108

Issue #92, opened 2022-04-09
Split XataRecord from Entity in codegen


SferaDev:
Right now:

```ts
export interface Foo extends XataRecord {
  property: string | null;
}
```

Improves DX:

```ts
export interface Foo {
  id: string;
  property: string | null;
}

export type FooRecord = Foo & XataRecord;
```

gimenete:
Why do you think splitting entities and records improves the DX?

SferaDev:
> Why do you think splitting entities and records improves the DX? Could you ellaborate on that?

I was working with xata on a side project, and when typing functions that received or returned the entities (without caring of the internal Record properties) was not easy. I was able to do ``Omit<Team, keyof XataRecord>`` but it wasn't idiomatic.

Let's say I have a function that creates many teams.

```ts
async function initTeams(teams: Team[]): Promise<TeamRecord[]>;
```

Without that differentiation you don't have an easy way to get the entity type (which you will probably pass around between functions).

gimenete:
I see. That makes sense 👍 Thanks for the explanation!

PR #98, opened 2022-04-12
Split entity from record in codegen

SferaDev:
Closes: #92


gimenete (commented):
LGTM. I'd like to see something in the README to explain the difference between entities and records.

SferaDev:
> LGTM. I'd like to see something in the README to explain the difference between entities and records.

I will update the readme today in another PR (adding more topics that are missing too)

gimenete (approved):
🚀

PR #99, opened 2022-04-12
Add inference for filter columns

SferaDev:
- Reduce subrequests for createMany
- Create serious-files-shave.md
- Split entity from record object
- Add inference magic to filter


SferaDev:
Closing in favor of #130

PR #107, opened 2022-04-19
Improve codegen error handling

gimenete:
When getting an ENOENT error, it may be because: the schema file cannot be read, or the output file cannot be written. The existing error in the second case is confusing, because the codegen treats it as the first case: telling the user there's no schema file when the problem is that the output file cannot be written.

This PR improves the error handling in these cases, and does some refactorings to be more DRY and have the code better organized.

Besides, the most common example when the file cannot be written is probably when the user specifies `generated/xata.ts` and `generated` doesn't exist (like we have in the README). I've made changes to create the directory if it doesn't exist.

### DIFF
--- a/codegen/package.json
+++ b/codegen/package.json
@@ -10,8 +10,8 @@
     "prebuild": "node scripts/add-version-file.js",
     "postbuild": "node scripts/inject-hashbang.js",
     "build": "tsc",
-    "example-ts": "npm run build && node --es-module-specifier-resolution=node dist generate example/xata -o example/xata.ts && sed -i '' 's|@xata.io/client|../../client/src|' example/xata.ts",
-    "example-js": "npm run build && node --es-module-specifier-resolution=node dist generate example/xata -o example/xata.js && sed -i '' 's|@xata.io/client|../../client/src|' example/xata.js"
+    "example-ts": "npm run build && node --es-module-specifier-resolution=node dist/cli.js generate example/xata -o example/xata.ts && sed -i '' 's|@xata.io/client|../../client/src|' example/xata.ts",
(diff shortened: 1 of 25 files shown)


codegen/package.json
```diff
@@ -10,8 +10,8 @@
     "prebuild": "node scripts/add-version-file.js",
     "postbuild": "node scripts/inject-hashbang.js",
     "build": "tsc",
-    "example-ts": "npm run build && node --es-module-specifier-resolution=node dist generate example/xata -o example/xata.ts && sed -i '' 's|@xata.io/client|../../client/src|' example/xata.ts",
-    "example-js": "npm run build && node --es-module-specifier-resolution=node dist generate example/xata -o example/xata.js && sed -i '' 's|@xata.io/client|../../client/src|' example/xata.js"
+    "example-ts": "npm run build && node --es-module-specifier-resolution=node dist/cli.js generate example/xata -o example/xata.ts && sed -i '' 's|@xata.io/client|../../client/src|' example/xata.ts",
+    "example-js": "npm run build && node --es-module-specifier-resolution=node dist/cli.js generate example/xata -o example/xata.js && sed -i '' 's|@xata.io/client|../../client/src|' example/xata.js"
```
gimenete, line 14: I think this broke at some point, because it was doing nothing for me.

SferaDev: Yes, I also noticed and it's solved the same way in https://github.com/xataio/client-ts/pull/98/files

SferaDev:
Something is still not working (erroring out when it shouldn't). Tested the commands shown in a newly created project without any file nor schema available.

image

gimenete:
@SferaDev that's weird. Good catch. I'm going to troubleshoot.

gimenete:
@SferaDev It should be fixed now. I also did some other minor changes, like adding a link to the readme.

SferaDev (approved):
Tested again and works fine!

PR #108, opened 2022-04-20
Expose pagination limits

SferaDev:
- Expose constants for pagination limits (size 200 and offset 800)
- Add integration tests that ensure those values are up to date
- Finally allow schema-less client (using proxies)


SferaDev:
What has just happened? This was a different branch and I merged base one but not this one...

PR #109, opened 2022-04-20
Expose pagination limits

SferaDev:
- Expose constants for pagination limits (size 200 and offset 800)
- Add integration tests that ensure those values are up to date
- Finally allow schema-less client (using proxies)


client/src/schema/repository.ts
```diff
@@ -321,6 +321,15 @@ export class BaseClient<D extends Record<string, Repository<any>>> {
 
     this.options = options;
     this.#links = links;
+
+    const factory = options.repositoryFactory || new RestRespositoryFactory();
+
+    this.db = new Proxy({} as D, {
+      get: (_target, prop) => {
+        if (typeof prop !== 'string') throw new Error('Invalid table name');
+        return factory.createRepository(this, prop);
```
SferaDev, line 330: I wasn't 100% sure about this ``this``, but since it's an arrow function that does not have ``this``, the parent one is persisted (still the base client).

gimenete (approved):
LGTM 👌

PR #111, opened 2022-04-20
Remove `generate` command

gimenete:
We just have one command and I don't think we will ever have another one. In any case we can add a different binary in the rare case that we need an additional command.

### DIFF
--- a/codegen/src/cli.ts
+++ b/codegen/src/cli.ts
@@ -51,7 +51,11 @@ program.parse();
 async function pullSchema() {
   spinner.warn('No local Xata schema found.');
   const { shouldUseCli } = await inquirer.prompt([
-    { name: 'shouldUseCli', message: 'Would you like to use the Xata CLI and clone a database?', type: 'confirm' }
+    {
+      name: 'shouldUseCli',
+      message: 'Would you like to use the Xata CLI to pull an existing database schema?',
+      type: 'confirm'
+    }
   ]);
 
   if (!shouldUseCli) {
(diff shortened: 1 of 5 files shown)


codegen/src/cli.ts
```diff
@@ -19,14 +19,10 @@ const spinner = ora();
...
-  .version(CODEGEN_VERSION);
-
-program
-  .command('generate')
-  .description('Generate code from a given Xata schema.')
+  .version(CODEGEN_VERSION)
   .argument(
     '[xata directory]',
-    `A path to your local Xata directory. If you don't have this, run the pull or \`init\` command on this CLI first.`,
+    `A path to your local Xata directory. If you don't have one, run 'xata init' or 'xata pull' first with the xata CLI https://github.com/xataio/cli`,
```
gimenete, line 25: Unrelated to the scope of the PR, but this was a quick improvement.
This CLI does not provide init or pull commands.

codegen/src/cli.ts
```diff
@@ -55,7 +51,11 @@ program.parse();
 async function pullSchema() {
   spinner.warn('No local Xata schema found.');
   const { shouldUseCli } = await inquirer.prompt([
-    { name: 'shouldUseCli', message: 'Would you like to use the Xata CLI and clone a database?', type: 'confirm' }
+    {
+      name: 'shouldUseCli',
+      message: 'Would you like to use the Xata CLI to pull an existing database schema?',
```
gimenete, line 56: Unrelated to the scope of the PR, but this was a quick improvement.

We don't use the term `clone` anywhere. Also, you are not creating a copy of a database, but just pulling the schema.

PR #112, opened 2022-04-20
Rename package to allow running through npx

SferaDev:
```
npx @xata.io/xata-codegen
```

image


SferaDev:
@gimenete Before merging we should ``deprecate`` old package and publish current version. I've cancelled the current build so the action does not attempt to publish the new one yet (approval on the rename).

PR #113, opened 2022-04-20
Use custom git token to trigger builds after push

gimenete:
When se use the GITHUB_TOKEN to push changes to git, other workflows are not triggered. By using a custom GIT_TOKEN we should avoid that.


SferaDev (approved):
We should do the same in frontend-next

PR #114, opened 2022-04-21
Run package on npx without rename

SferaDev:
- Rename bin executable
- Add changeset
- Rename package
- Revert "Rename package"
- Export two different bin executables


gimenete (approved):
🚀

PR #115, opened 2022-04-22
Make Selectable not "Identifiable"

gimenete:
I think it makes sense that Selectable doesn't extend from Identifiable. For all these methods the Selectable part is not expected to contain an id:

```ts
  abstract create(object: Selectable<T>): Promise<T>;
  abstract createMany(objects: Selectable<T>[]): Promise<T[]>;
  abstract insert(id: string, object: Selectable<T>): Promise<T>;
  abstract update(id: string, object: Partial<Selectable<T>>): Promise<T>;
  abstract updateOrInsert(id: string, object: Selectable<T>): Promise<T>;
```

Because either we specify an `id` as an argument, or the user can't specify an id (e.g. create).

If we need to make an argument Selectable and Identifiable let's make that specific to that case.

### DIFF
--- a/client/src/schema/repository.ts
+++ b/client/src/schema/repository.ts
@@ -156,7 +156,7 @@ export class RestRepository<T extends XataRecord> extends Repository<T> {
       ...fetchProps
     });
 
-    const finalObjects = await this.any(...response.recordIDs.map((id) => this.filter('id', id))).getAll();
+    const finalObjects = await this.any(...response.recordIDs.map((id) => this.filter('id' as any, id))).getAll();
     if (finalObjects.length !== objects.length) {
       throw new Error('The server failed to save some records');
     }
--- a/client/src/schema/selection.ts
+++ b/client/src/schema/selection.ts
@@ -1,27 +1,13 @@
 import { XataRecord } from '..';
 import { StringKeys, UnionToIntersection, Values } from '../util/types';
 import { Query } from './query';
-import { Identifiable } from './record';
 
 type Queries<T> = {
   [key in keyof T as T[key] extends Query<any> ? key : never]: T[key];
 };
 
-type OmitQueries<T> = {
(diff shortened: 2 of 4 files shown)


client/src/schema/selection.ts
```diff
@@ -1,27 +1,13 @@
...
-
-type OmitLinks<T> = {
-  [key in keyof T as T[key] extends XataRecord ? never : key]: T[key];
-};
-
-type OmitMethods<T> = {
-  // eslint-disable-next-line @typescript-eslint/ban-types
-  [key in keyof T as T[key] extends Function ? never : key]: T[key];
-};
-
```
gimenete, line 22: I deleted these because were not being used.

client/src/schema/repository.ts
```diff
@@ -156,7 +156,7 @@ export class RestRepository<T extends XataRecord> extends Repository<T> {
       ...fetchProps
     });
 
-    const finalObjects = await this.any(...response.recordIDs.map((id) => this.filter('id', id))).getAll();
+    const finalObjects = await this.any(...response.recordIDs.map((id) => this.filter('id' as any, id))).getAll();
```
SferaDev, line 159: This actually is an error. Since id always exists it can be filtered, sorted, etc...

client/src/schema/selection.ts
```diff
@@ -1,27 +1,13 @@
...
 type InternalProperties = keyof XataRecord;
-export type Selectable<T extends XataRecord> = Omit<T, InternalProperties> & Identifiable;
+export type Selectable<T extends XataRecord> = Omit<T, InternalProperties>;
```
SferaDev, line 10: What about Partial<Identifiable>

SferaDev (approved):

gimenete:
Closing based on a conversation with @SferaDev to solve this problem in a different way.

PR #116, opened 2022-04-22
Prefer hooks over installing the CLI

gimenete:
As we discussed in [this proposal](https://www.notion.so/How-Did-Hercules-Kill-the-Hydra-or-How-We-Can-Kill-the-Two-Headed-Monster-f1a7baa50f6c4bd082e9a666af3a747f) we are going to make the xata cli have [hooks and install the codegenerator](https://github.com/xataio/cli/pull/51). So we no longer need the other way around. This PR clean ups the code that downloads the CLI and also does some refactorings to avoid having too many files with just one function.


SferaDev (approved):

SferaDev:
@gimenete I would merge this and tag a release with what we have. And leave the update dependencies for the next one.

gimenete:
@SferaDev sounds good to me!

Issue #121, opened 2022-04-26
Fix circular references


SferaDev:
<img width="980" alt="Screenshot 2022-04-26 at 10 58 50" src="https://user-images.githubusercontent.com/2181866/165263243-f5d6fce7-1e0f-44ee-81a0-2e9420a45865.png">

SferaDev:
In theory, coming to TS 3.7: https://github.com/microsoft/TypeScript/pull/33050.

PR #122, opened 2022-04-26
Fix circular references and implement link reading

SferaDev:
Closes: #121

Fixes:

- Circular dependency type instantiation (TypeScript)
- Links are now loaded until a given depth (10)
- Below the given depth, they're a ``XataRecord`` that can be ``read()``
- Fix ``update(data)`` not properly handling links


codegen/src/codegen.ts
```diff
@@ -20,10 +20,16 @@ function getTypeName(tableName: string) {
...
+  const properties = [
+    ...columns.map((column) => generateColumnType(column)),
+    ...revLinks.map((link) => `${link.table}: Query<${getTypeName(link.table)}>`)
+  ];
+
   return `export interface ${getTypeName(table.name)} {
-    ${columns.map((column) => generateColumnType(column)).join('\n')}
-    ${revLinks.map((link) => `${link.table}: Query<${getTypeName(link.table)}>`).join('\n')}
+    ${properties.join('\n')}
+    [key: string]: unknown;
```
gimenete, line 32: Does this solve the circular references?

The problem is that now users won't have compilation errors if they do `obj.columnThatDoesNotExist`, right?

SferaDev: > Does this solve the circular references?

Yes, it's a TS long living error (they attempted to fix it in 3.7 but you still need a named index access to make them work)

> The problem is that now users won't have compilation errors if they do obj.columnThatDoesNotExist, right?

That's right, it will return "unknown" type but no compile error.

That's why utils like zod have ``lazy`` initializers and sometimes they're unable to type themselves with circular references. It can be solved but with a tradeoff.

gimenete (dismissed):

SferaDev:
Let's hold on merging this. I have been thinking and since the circular recursion is coming from the conditional types I can theoretically solve it with a depth limit (ie. N levels and ignore rest)

client/src/schema/repository.ts
```diff
@@ -352,17 +356,26 @@ export class BaseClient<D extends Record<string, Repository<any>> = Record<strin
...
+            },
+            update: (data: any) => {
+              return this.db[linkTable].update(id, data);
+            },
+            delete: () => {
+              return this.db[linkTable].delete(id);
             }
           };
+        } else if (id) {
+          const item = await this.db[linkTable].read(id, { ignoreLinks: true });
```
gimenete, line 377: This is making a request to the server, right? This would cause an N+1 problem. We should just `initObject` but not make a request to the server. Even if we just have the id.

client/src/schema/repository.ts
```diff
@@ -352,17 +356,26 @@ export class BaseClient<D extends Record<string, Repository<any>> = Record<strin
...
-        } else if (id) {
+      if (isIdentifiable(value)) {
+        const { id } = value;
+
+        if (depth === -1 || depth >= 10) {
           o[field] = {
             id,
-            get: () => {
-              this.db[linkTable].read(id);
+            xata: { version: 1 },
```
gimenete, line 365: 🤔  is there a case when we don’t get the version? If that’s the case I’d prefer to have here an undefined or if the field needs to be defined we can default to -1. But first, I’d like to understand why in this case we don’t get the version.

client/src/schema/repository.ts
SferaDev: It will make up to 10 requests (max depth).

If we only initObject, all properties except ``id`` will always be undefined (which is not what we want)

client/src/schema/repository.ts
SferaDev: This is the depth >= 10 case, where we have already gone too deep, and we do not retrieve from the server all the properties and leave the id + the read/update/delete methods

client/src/schema/repository.ts
SferaDev: You can take a look at the integration test I have added (has an infinite recursive depedency) that ensures we don't have N + 1 calls here.

client/src/schema/repository.ts
SferaDev:
```suggestion
            xata: { version: -1 },
```

SferaDev:
Closing in favor of https://github.com/xataio/client-ts/pull/130

Issue #124, opened 2022-04-26
Codegen with tables only consisting of numbers


SferaDev:
<img width="576" alt="Screenshot 2022-04-26 at 13 34 14" src="https://user-images.githubusercontent.com/2181866/165291365-68470d37-d3c6-42b7-92f5-7bd546c0d752.png">
<img width="668" alt="Screenshot 2022-04-26 at 13 34 52" src="https://user-images.githubusercontent.com/2181866/165291374-3f216da4-75c8-469d-b0b8-5f8d8054ba95.png">

PR #126, opened 2022-04-27
Allow table names to start with a number

SferaDev:
Closes: #124


gimenete (approved):
Good idea prefixing with $ 👍

codegen/src/codegen.ts
```diff
@@ -15,6 +15,10 @@ export type GenerateOptions = {
 
 function getTypeName(tableName: string) {
   const snglr = pluralize.singular(tableName);
+
+  // If table starts with a number, prepend $ sign
```
SferaDev, line 19:
```suggestion
  // If table starts with a number, prepend a $ sign
```

PR #127, opened 2022-04-28
New importer package

gimenete:
This PR adds a new package. Maybe there are a few tweaks to do it but I think it's in a good shape to be used. More information on how to use it in the README and in the command's `--help`.

The main goal in this PR was to make a CLI app. I've tried to make the core functionality abstracted enough so it can be used as an API, but I haven't tried that yet. I'll do another PR for further changes regarding that.

Right now the only format supported is CSV, but similarely I've tried to abstract things enough to support more file formats in the future.

This PR supports creating a table if it doesn't exist as well as updating an existing one. It also supports creating new columns in an existing columns if neccessary. It tries to guess the right data type for each column, etc.


.changeset/config.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://unpkg.com/@changesets/config@1.7.0/schema.json",
   "changelog": "@changesets/cli/changelog",
   "commit": false,
-  "linked": [["@xata.io/client", "@xata.io/codegen"]],
+  "linked": [["@xata.io/client", "@xata.io/codegen", "@xata.io/importer"]],
```
SferaDev, line 5: Not sure if we want to link all dependency versions. The codegen kind of made sense, but maybe no need to bump all minor/major changes for all of them at the same time.

importer/src/cli.ts
```diff
@@ -0,0 +1,21 @@
...
+  .version(IMPORTER_VERSION)
+  .argument('<file>', 'File containing the data to import. Pass "-" to pipe content instead.')
+  .option('--table <table>', 'The table in which to import the data.')
+  .option('--columns <columns>', 'List of column names.')
+  .option('--types <types>', 'List of column names.')
+  .option('--noheader', 'Specify that the CSV file has no header.')
+  .option('--format <format>', 'The format of the input source.', 'csv')
+  .option('--create', "Whether the table or columns should be created if they don't exist.", false)
+  .option('--force', 'Whather confirmation should be asked when creating the table or columns.', false)
+  .option('--branch', 'The branch to import data to.', 'main')
```
SferaDev, line 17:
```suggestion
  .option('--branch <branch>', 'The branch to import data to.', 'main')
```

importer/src/cli.ts
```diff
@@ -0,0 +1,21 @@
...
+  .name('xata-importer')
+  .description('Importer utility for xata.io')
+  .version(IMPORTER_VERSION)
+  .argument('<file>', 'File containing the data to import. Pass "-" to pipe content instead.')
+  .option('--table <table>', 'The table in which to import the data.')
+  .option('--columns <columns>', 'List of column names.')
+  .option('--types <types>', 'List of column names.')
+  .option('--noheader', 'Specify that the CSV file has no header.')
+  .option('--format <format>', 'The format of the input source.', 'csv')
+  .option('--create', "Whether the table or columns should be created if they don't exist.", false)
```
SferaDev, line 15: Not sure I get the ``--create`` flag. If I don't pass it, it errors me out. If I do pass it still asks me for confirmation.

In case you don't pass it I would ask interactively and have the flag to not ask.

Screenshot 2022-05-04 at 08 20 19

importer/src/cli.ts
```diff
@@ -0,0 +1,21 @@
...
+
+program
+  .name('xata-importer')
+  .description('Importer utility for xata.io')
+  .version(IMPORTER_VERSION)
+  .argument('<file>', 'File containing the data to import. Pass "-" to pipe content instead.')
+  .option('--table <table>', 'The table in which to import the data.')
+  .option('--columns <columns>', 'List of column names.')
+  .option('--types <types>', 'List of column names.')
+  .option('--noheader', 'Specify that the CSV file has no header.')
```
SferaDev, line 13: I would have expected with this flag to use the existing table columns, but I guess order cannot be guaranteed. Could we make it interactive somehow? [just a comment, but I understand this as a future improvement]

SferaDev (commented):
I would like to add new rows to an already existing table/schema. However it's not properly comparing the types from the online schema.

Screenshot 2022-05-04 at 08 44 48

Screenshot 2022-05-04 at 08 45 03

Screenshot 2022-05-04 at 08 29 07

SferaDev:
Since we rely on the client we also have the createMany error handling issue

Screenshot 2022-05-04 at 08 31 36

giphy (1)

importer/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "bugs": {
+    "url": "https://github.com/xataio/client-ts/issues"
+  },
+  "homepage": "https://github.com/xataio/client-ts/blob/main/importer/README.md",
+  "dependencies": {
+    "@xata.io/client": "^0.5.0",
+    "camelcase": "^6.3.0",
+    "commander": "^9.2.0",
+    "cross-fetch": "^3.1.5",
+    "csvtojson": "^2.0.10",
```
SferaDev, line 30: Not 100% sure about this dependency, seems not actively maintained

Screenshot 2022-05-04 at 08 37 34

importer/src/cli.ts
```diff
@@ -0,0 +1,21 @@
+import { program } from 'commander';
+import { run } from './cli-run.js';
+import { IMPORTER_VERSION } from './version.js';
+
+program
+  .name('xata-importer')
+  .description('Importer utility for xata.io')
+  .version(IMPORTER_VERSION)
+  .argument('<file>', 'File containing the data to import. Pass "-" to pipe content instead.')
```
SferaDev, line 9: I would add another option for ``delimiter`` and probably attempt to make it "smart".

Although CSV stands for comma-separated-values, the delimiter could be anything. Most popular are semicolons, tabs, spaces, etc...

importer/src/csv.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Converter } from 'csvtojson/v2/Converter';
+import internal from 'stream';
+import { ParseOptions } from './index.js';
+
+export async function parseString(text: string, options: ParseOptions) {
+  return process(initConverter(options).fromString(text), options);
+}
+
+export async function parseFile(file: string, options: ParseOptions) {
+  return process(initConverter(options).fromFile(file), options);
```
SferaDev, line 11: Do we handle different encodings? I wouldn't assume UTF-8 by default, and maybe make it configurable

SferaDev (commented):
@gimenete Works really good, I love the type inference 💯 

Added a lot of comments, most of them can be taken care after merging. I'll leave you decide how you want to prioritise them. Ping me if you need an approval.

importer/src/cli.ts
gimenete: Good idea.

Question about one use case for that. How could we specify tab delimiters with flags in the CLI? 🤔

importer/src/csv.ts
gimenete: Good suggestion. I'm not sure if the csv parser already takes care of that somehow (checking if there's a BOM for UTF), but I guess for most cases if the encoding is a different one, we'd need a way to specify it.

gimenete:
> Since we rely on the client we also have the createMany error handling issue

Yup. I'll upgrade it once we release a new version of the client.

> I would like to add new rows to an already existing table/schema. However it's not properly comparing the types from the online schema. 

I'll look into that.

Thanks for your review!! 🙏

importer/src/cli.ts
SferaDev: What about "\t"

importer/src/cli.ts
SferaDev: Yeah, although I would make it always ask and maybe have a interactive/non-interactive flag that disables inquirer everywhere and fails if the flags for ``create`` or ``force`` are not provided. WDYT?

gimenete:
@SferaDev the problem with the email/string thing should be fixed. I was not implementing the easiest case for `castType`: when the types are equal 🤦‍♂️ 

I've addressed some feedback and I think the rest can be implemented or explored in other PRs.

SferaDev (approved):
Tested again, now it works properly.

I would create several follow-up issues:

- Add delimiter option (semicolons, tabs, spaces)
- Review different encodings (latin1)
- Add a non-interactive flag, make all commands to inquiry instead of erroring out, if flags are provided ignore the inquiry question (related to https://github.com/xataio/client-ts/pull/127/files#r864487665)
- Add a way to call interactively as an npm package (API)

gimenete:
Thanks @SferaDev for the review and the feedback! I've created https://github.com/xataio/client-ts/issues/142

Issue #9, opened 2022-02-14
Inference for nested columns properties (select, filter...)


SferaDev:
Even though I'm still a couple of weeks from onboarding, I had some curiosity about the state of the TS client, and I took a look to the code. I've noticed that nested property access from string was not possible yet (even if the API supports it).

When [string literal types ](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) became a thing on TS I experimented a bit with them and I think they can be useful here. I've quickly adapted a small playground I had from a few months ago to something similar to what the API supports for selects and filters. Just leaving it here so you can take a look.

```typescript
type StringKeys<O> = Extract<keyof O, string>;
type Values<O> = O[keyof O];

export type ObjectPaths<O> = O extends Array<unknown>
  ? never // Not sure about how the API will handle arrays of elements (multiple field type seems to be string only and it's not clear if it will be generic)
  : O extends Record<string, any>
  ?
      | "*"
      | Values<{
          [K in StringKeys<O>]: O[K] extends Record<string, any>
            ? `${K}.${ObjectPaths<O[K]>}`
            : K;
        }>
  : "";

export type ObjectProperty<O, P extends ObjectPaths<O>> = P extends "*"
  ? Values<O>
  : P extends keyof O
  ? O[P]
  : P extends `${infer K}.${infer V}`
  ? K extends keyof O
    ? V extends ObjectPaths<O[K]>
      ? ObjectProperty<O[K], V>
      : never
    : never
  : never;

interface Example1 {
  foo: boolean;
  bar: { a: number; b: string };
  baz: Array<{ c: string[] }>;
}

type Example2 = {
  foo: boolean;
  bar: { a: number; b: string };
  baz: Array<{ c: string[] }>;
};

type Paths1 = ObjectPaths<Example1>;
type Paths2 = ObjectPaths<Example2>;

const path1: Paths1 = "*";
const path2: Paths2 = "*";
const path3: Paths1 = "foo";
const path4: Paths2 = "foo";
const path5: Paths1 = "bar.a";
const path6: Paths2 = "bar.a";
const path7: Paths1 = "bar.*";
const path8: Paths2 = "bar.*";
// @ts-expect-error
const path9: Paths1 = "unknown";
// @ts-expect-error
const path10: Paths2 = "unknown";

const test1A: ObjectProperty<Example1, "foo"> = true;
const test1B: ObjectProperty<Example2, "foo"> = false;

[...]

PR #130, opened 2022-04-28
Selectable columns type inference

SferaDev:
Closes: #9 

This is a huge PR and possibly one of the last blockers we had before recommending users to switch to our TS SDK.


client/src/schema/selection.ts
```diff
@@ -1,25 +1,53 @@
...
+  ? O
+  : Pick<O, Extract<Key[number], keyof O>> &
+      (IncludeXataRecord extends true ? XataRecord : unknown) &
+      UnionToIntersection<
+        Values<{
+          [K in Exclude<Key[number], keyof O>]: K extends `${infer N}.${infer M}`
+            ? N extends keyof O
+              ? {
+                  [key in N]: M extends '*'
+                    ? O[N]
```
SferaDev, line 33: Probably duplicated, add tests

SferaDev:
Ready for review.

There's one scenario not fully finished: Nullable forward. Right now it's not really important as we don't have support for that on the schema/codegen. But I want to review it later on. Opening an issue so I don't forget: https://github.com/xataio/client-ts/issues/141

client/src/util/types.ts
```diff
@@ -1,4 +1,16 @@
...
+export type Values<O> = O[StringKeys<O>];
 
 export type UnionToIntersection<T> = (T extends any ? (x: T) => any : never) extends (x: infer R) => any ? R : never;
+
+export type If<Condition, Then, Else> = Condition extends true ? Then : Else;
+
+export type IsObject<T> = T extends Record<string, any> ? true : false;
+export type IsArray<T> = T extends Array<any> ? true : false;
+
+export type NonEmptyArray<T> = T[] & { 0: T };
```
gimenete, line 12: Nice. Didn't know this trick!

codegen/example/xata.js
```diff
@@ -9,7 +9,7 @@ import { BaseClient, RestRespositoryFactory } from "../../client/src";
  * @property {() => Promise<void>} delete
  * @property {string=} name
  * @property {string[]=} labels
- * @property {User=} owner
+ * @property {UserRecord=} owner
```
gimenete, line 12: Not for this PR, but I wonder if we could generate a .d.ts when generating JS code and importing these types instead of using `@property` annotations. This is what we talked a few weeks ago and forgot to open an issue. I'll create one.

gimenete (approved):
Excellent job!! 👏 👏 👏 👏 

I haven't tested the library in any code, but will try to do that tomorrow. I don't want to block the PR and the changes look good, so I'm approving.

PR #133, opened 2022-04-30
Add eslint rule: no floating promises

gimenete:
Also, I've fixed the `npm run lint` and `npm run lint:fix` npm scripts.

### DIFF
--- a/client/.eslintrc.cjs
+++ b/client/.eslintrc.cjs
@@ -1,0 +1,13 @@
+module.exports = {
+  ignorePatterns: ["dist"],
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+    project: 'client/tsconfig.json'
+  },
+  rules: {
+    '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/ban-types': 'off',
+    '@typescript-eslint/no-floating-promises': 'error',
+  }
+};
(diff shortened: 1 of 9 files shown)


client/.eslintrc.cjs
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  ignorePatterns: ["dist"],
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+    project: 'client/tsconfig.json'
```
SferaDev, line 6: Not sure this is taking the config file properly

codegen/.eslintrc.cjs
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  ignorePatterns: ["dist", "example", "scripts"],
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+    project: 'codegen/tsconfig.json'
```
SferaDev, line 6: Same

client/.eslintrc.cjs
gimenete: 🤔 this is correct to run eslint from the root dir. Like in `eslint .`

How are you trying to run eslint? From the `client` dir in this case?

SferaDev: Yup, as expected

Screenshot 2022-05-02 at 12 19 55

SferaDev (approved):

PR #136, opened 2022-05-02
Rename create and insert

SferaDev:
```ts
  /*
   * Creates a single record in the table.
   * @param object Object containing the column names with their values to be stored in the table.
   * @returns The full persisted record.
   */
  abstract create(object: Data & Partial<Identifiable>): Promise<Record>;

  /**
   * Creates a single record in the table with a unique id.
   * @param id The unique id.
   * @param object Object containing the column names with their values to be stored in the table.
   * @returns The full persisted record.
   */
  abstract create(id: string, object: Data): Promise<Record>;
```


client/src/schema/repository.ts
```diff
@@ -132,7 +133,23 @@ export class RestRepository<Data extends BaseData, Record extends XataRecord = D
...
 
-  async create(object: Data): Promise<Record> {
+  async create(object: Data): Promise<Record>;
+  async create(recordId: string, object: Data): Promise<Record>;
+  async create(a: string | Data, b?: Data): Promise<Record> {
+    if (isString(a) && isObject(b)) {
+      if (a === '') throw new Error("The id can't be empty");
+      return this.#insertRecordWithId(a, b);
+    } else if (isObject(a) && isString(a.id)) {
+      if (a.id === '') throw new Error("The id can't be empty");
```
gimenete, line 143:
```suggestion
      if (!a.id) throw new Error("The id can't be empty, null or undefined");
```

The test will need to be updated.

test/integration.test.ts
```diff
@@ -532,3 +460,104 @@ describe('integration tests', () => {
...
+  });
+
+  test('create user with empty inline id is not allowed', async () => {
+    expect(
+      client.db.users.create({
+        id: '',
+        full_name: 'John Doe 3',
+        email: 'john3@doe.com'
+      })
+    ).rejects.toMatchInlineSnapshot(`[Error: The id can't be empty]`);
```
gimenete, line 551:
```suggestion
    ).rejects.toMatchInlineSnapshot(`[Error: The id can't be empty, null or undefined]`);
```

gimenete (approved):

client/src/schema/repository.ts
SferaDev: `isString(a.id)` checks for null and undefined already

test/integration.test.ts
SferaDev: If null or undefined the error "[Error: Invalid arguments for create method]". We already have a test for it: https://github.com/xataio/client-ts/pull/136/files#diff-8fdaf3b73054c861d4b1afc72e76086d85c318c75bde88a452ef344c24b7d8b0R554-R563

client/src/schema/repository.ts
gimenete: Cool

PR #146, opened 2022-05-05
Filter types with type inference and safety

SferaDev:
- Build type definitions for what the API exposes on ``filters``
- Add type safety tests for valid/invalid semantics
- Update the existing operators to match the new types

Future improvements (outside this PR): Created #153

- Add some missing operators
- Add tests for all operators
- Add some more integration tests


client/src/util/lang.ts
```diff
@@ -13,7 +13,7 @@ export function compactObject<T>(obj: Record<string, T | null | undefined>): Rec
 export type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
 
 export function isObject(value: any): value is object {
-  return value !== undefined && value !== null && typeof value === 'object';
+  return value !== undefined && value !== null && typeof value === 'object' && !Array.isArray(value);
```
gimenete, line 16: Nitpick. Maybe we can do this expression sorter:

```suggestion
  return value != null && typeof value === 'object' && !Array.isArray(value);
```

gimenete (approved):
There's a lot to process here 😅  but I've taken a few looks and 👍 

I just left a nitpick suggestion.

Btw, the spec tests are great. Not just for making sure we don't break anything, but as a source of documentation (sort of 😆 ):

client/src/util/lang.ts
SferaDev: I would rather do:

```suggestion
  return Boolean(value) && typeof value === 'object' && !Array.isArray(value);
```

SferaDev:
> There's a lot to process here 😅 but I've taken a few looks and 👍

Yeah, sorry for that... It's easier to write it than to review/understand it. There's a lot of things happening in these types...

gimenete:
No worries. I don't have so deeper understanding of complex typescript types 😅 That's the thing. I'm not saying is more complex than it should be. In fact, I think our types are much simpler than the types of similar tools. Great job!

SferaDev:
> I don't have so deeper understanding of complex typescript types 😅 

It's an acquired taste 🤣 

> I'm not saying is more complex than it should be. In fact, I think our types are much simpler than the types of similar tools. 

Agreed, most of the existing tools and their type definitions were done before template literals existed and rely on a lot of hacks and internal ``as any``. We can proudly say that we only have it in one place (and I have a pending task to build a repro and report it as a bug).

PR #147, opened 2022-05-06
Database branch resolution

gimenete:
As we discussed we want to use the following logic to calculate the current branch:

- If the env variable `XATA_BRANCH` exists, use it.
- Otherwise, calculate the git branch, check if there's a database branch with the same name.
- If it exists, use it.
- If no database name exists with the same name, use `main` (until the server implements a default branch concept). This can be a dangerous if by mistake a developer connects to the production database. This will be fixed with the default/development branch concept and having scopes in API keys to prevent development API keys to have access to production. In the meantime, since we are in private beta, I think that's fine for now. Another option is to add an option to XataOptions to specify the name of the default branch. But, I prefer the logic to care care of that so we can have a constructor without any arguments: `new XataClient()`.

Getting environment variables and calculating the git branch is platform-specific code. I've implemented it for Node and Deno.


client/src/util/fetch.ts
```diff
@@ -0,0 +1,13 @@
+import { FetchImpl } from '../api/fetcher';
+
+export function getFetchImplementation(fetch?: FetchImpl) {
+  const globalFetch = typeof fetch !== 'undefined' ? fetch : undefined;
+  const fetchImpl = fetch ?? globalFetch;
```
SferaDev, line 5:
```suggestion
export function getFetchImplementation(userFetch?: FetchImpl) {
  const globalFetch = typeof fetch !== 'undefined' ? fetch : undefined;
  const fetchImpl = userFetch ?? globalFetch;
```

SferaDev:
> Do we want to stop exposing the branch-strategy thing from users? I mean, leaving branch as an optional parameter that only accepts a string.

I'd give as much as possibilities to the users (also because we are users too). For example, usage of the sdk in browser environments might not resolve any branch, and I'd like to have a way to resolve the branch manually 😄

gimenete:
> For example, usage of the sdk in browser environments might not resolve any branch, and I'd like to have a way to resolve the branch manually 😄

That's a good point 👍 !

SferaDev:
What about adding deployment providers git branch guessing?

Ie:

VERCEL_GIT_COMMIT_REF for Vercel
BRANCH for Netlify
CF_PAGES_BRANCH for Cloudflare Pages

gimenete:
> What about adding deployment providers git branch guessing?

I was reluctant to that, because I didn't want to introduce vendor-specific stuff in our code, but I think it makes sense and it's going to make things much easier for our users. I'll add that!

client/src/schema/config.ts
```diff
@@ -0,0 +1,71 @@
...
+  'BRANCH' // Netlify. Putting it the last one because it is more ambiguous
+];
+
+export async function getBranch(fetchProps: Omit<FetcherExtraProps, 'workspacesApiUrl'>): Promise<string | undefined> {
+  try {
+    const env = getBranchByEnvVariable();
+    if (env) return env;
+  } catch (err) {
+    // Ignore
+  }
```
SferaDev, line 19: [nitpick] Would probably move the try/catch to ``getBranchByEnvVariable``. It's not clear why we should catch, so easy to remove during a refactor.

client/src/types/global-branch.d.ts
```diff
@@ -0,0 +1,3 @@
+declare const XATA_BRANCH: string | undefined;
+declare const VERCEL_GIT_COMMIT_REF: string | undefined; // Vercel
+declare const XATA_BRANCH: string | undefined;
```
SferaDev, line 3:
```suggestion
```

Old?

client/src/schema/config.ts
gimenete: Makes sense!

client/src/types/global-branch.d.ts
```diff
@@ -0,0 +1,3 @@
...
```
SferaDev, line 1: Probably move to ``client/src/types/global-cloudflare.d.ts`` as globals are only a cloudflare thing

gimenete: Yep. I got distracted and I thought I didn't made these changes. I'll put it back in the cloudflare globals 👍

client/src/types/global-cloudflare.d.ts
```diff
@@ -0,0 +1,2 @@
+declare const XATA_DATABASE_URL: string | undefined;
+declare const XATA_API_KEY: string | undefined;
```
SferaDev, line 2: [question] Probably rename to something like ``global-variables.d.ts`` as other vendors other than CF could use the same approach?

codegen/src/codegen.ts
```diff
@@ -123,10 +123,10 @@ export async function generate({
     export class XataClient extends BaseClient<{
       ${tables.map((table) => `"${table.name}": Repository<${getTypeName(table.name)}>;`).join('\n')}
     }> {
-      constructor(options: XataClientOptions) {
+      constructor(options?: Partial<XataClientOptions>) {
```
SferaDev, line 126:
```suggestion
      constructor(options?: XataClientOptions) {
```

Shouldn't all properties be now optional?

gimenete: The fields `branch` and `apiKey` are required in our internal code. The user may or may not specify them, but we will calculate values for them. So, either we put all as optional and internally we have a different type or we use `options?.apiKey` instead of `options.apiKey`.

SferaDev: Yeah, don't see why we can't make everything optional. Proposal: https://github.com/xataio/client-ts/pull/165

SferaDev (approved):
LGTM!

I will test it further with real code examples (in the playground for example), but I don't see why we can't merge it as-is.

PR #149, opened 2022-05-06
Improve build sort object helper

SferaDev:
- Add API type
- Add tests for API type
- Improve helper that builds the sort object to persist order


gimenete (approved):
Nice! ✨

PR #150, opened 2022-05-06
Add support for all bulk operations

SferaDev:
- Add support for all bulk operations (update, delete, create, createOrUpdate)
- Add a warning from 100s of records +
- We will update when the bulk endpoints are implemented


client/src/schema/repository.ts
```diff
@@ -235,7 +342,37 @@ export class RestRepository<Data extends BaseData, Record extends XataRecord = D
...
+  async createOrUpdate(object: EditableData<Data>): Promise<SelectedPick<Record, ['*']>>;
+  async createOrUpdate(recordId: string, object: EditableData<Data>): Promise<SelectedPick<Record, ['*']>>;
+  async createOrUpdate(objects: EditableData<Data>[]): Promise<SelectedPick<Record, ['*']>[]>;
+  async createOrUpdate(
+    a: string | EditableData<Data> | EditableData<Data>[],
+    b?: EditableData<Data>
+  ): Promise<SelectedPick<Record, ['*']> | SelectedPick<Record, ['*']>[]> {
+    // Create or update many records
+    if (Array.isArray(a)) {
+      if (a.length > 100) {
```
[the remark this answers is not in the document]
SferaDev: Don't think so, I did a quick test and created 100k records in a single call (took around 50 seconds).

The warning is only temporary until we have backend support (each operation is a network call right now, huge overhead)

.changeset/cool-ants-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xata.io/client': patch
```
SferaDev, line 2:
```suggestion
'@xata.io/client': minor
```

SferaDev: We're renaming ``createMany`` so a breaking change in a 0.x version.

Issue #157, opened 2022-05-07
Review if we are incorrectly extending XataRecord on nested objects


SferaDev:
https://github.com/xataio/client-ts/blob/f46df88912b2bac2ddae598d9d08efc1f44b00be/client/src/schema/selection.ts#L77

Issue #156, opened 2022-05-07
Review inference for nested objects that are not links


SferaDev:
https://github.com/xataio/client-ts/blob/f46df88912b2bac2ddae598d9d08efc1f44b00be/client/src/schema/selection.ts#L58

If typescript does not find circular references this should not have a depth check

PR #158, opened 2022-05-09
Improve nullable and record selection

SferaDev:
Closes: #156
Closes: #157

Add more tests to ensure these scenarios are always checked

### DIFF
--- a/client/src/schema/selection.ts
+++ b/client/src/schema/selection.ts
@@ -74,20 +74,24 @@ type NestedValueAtColumn<O, Key extends SelectableColumn<O>> =
     ? N extends DataProps<O>
       ? {
           [K in N]: M extends SelectableColumn<NonNullable<O[K]>>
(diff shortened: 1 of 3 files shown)


client/src/schema/selection.ts
```diff
@@ -74,20 +74,24 @@ type NestedValueAtColumn<O, Key extends SelectableColumn<O>> =
     ? N extends DataProps<O>
       ? {
           [K in N]: M extends SelectableColumn<NonNullable<O[K]>>
-            ? NestedValueAtColumn<NonNullable<O[K]>, M> & XataRecord
+            ? NonNullable<O[K]> extends XataRecord
```
SferaDev, line 77: This new check adds ``XataRecord`` only for items that are records (not object types). Re #157

gimenete (approved):
👏  Great job! Is this perfection already? 😜

Issue #152, opened 2022-05-06
Improve error handling


gimenete:
A couple of things that I found while dog-fooding:

- Errors returned by `withStatus` don't have a stack trace. I had an error and it was hard to know where it was coming from, because the backend error was very generic.
- The bulk API sometimes returns errors without a message: e.g. `{ status: 409 },` so we finally get an `Unknown error`. We should ignore those and get the error messages from the rest of operations.

SferaDev:
Merged: https://github.com/xataio/xata/pull/710

PR #163, opened 2022-05-10
Include stack trace with errors

SferaDev:
Closes: https://github.com/xataio/client-ts/issues/152


test/integration.test.ts
```diff
@@ -461,12 +476,37 @@ describe('integration tests', () => {
...
 
-    // @ts-expect-error
-    expect(client.db.users.createMany(invalidUsers)).rejects.toMatchInlineSnapshot(`
+    expect(client.db.users.createMany(invalidUsers)).rejects.toMatchInlineSnapshot(
+      {
+        message: 'Unknown error',
+        status: 400,
+        errors: [
+          {
+            //message: 'record ignored: transaction failed',
```
SferaDev, line 487: To be updated when change in production is deployed

gimenete (approved):
LGTM!

Only downside of this solution I can see is that you cannot do `instanceof Error`. Not sure if that's a big deal. WDYT

gimenete:
> The bad thing about the Error class instance is that it only has string messages.

I guess we could extend the Error class. WDYT?

gimenete:
@SferaDev looks great!

PR #166, opened 2022-05-11
Add search in xata client

SferaDev:
Closes: #164


client/src/schema/repository.ts
```diff
@@ -517,12 +539,25 @@ export class BaseClient<D extends Record<string, Repository<any>> = Record<strin
...
       }
     });
   }
 
+  async search<Tables extends keyof D>(
+    query: string,
+    tables: Tables[] = Object.keys(this.db) as Tables[],
+    options?: { fuzziness?: number }
+  ): Promise<{ [Model in GetArrayInnerType<typeof tables>]: Awaited<ReturnType<D[Model]['search']>> }> {
+    // TODO: Implement global search with a single call, REST repository abstraction needed
```
SferaDev, line 553: We need to revisit this, but a discussion on plugins/namespaces should happen before. We can discuss it in our next 1:1 or a team sync.

test/integration.test.ts
```diff
@@ -817,3 +824,44 @@ describe('getBranch', () => {
...
+
+    expect(users[0].id).toBeDefined();
+    expect(users[0].read).toBeDefined();
+    expect(users[0].full_name?.includes('fruits')).toBeTruthy();
+
+    expect(teams[0].id).toBeDefined();
+    expect(teams[0].read).toBeDefined();
+    expect(teams[0].name?.includes('fruits')).toBeTruthy();
+  });
+});
```
SferaDev, line 867: This randomly fails (indexes not created yet, API returns empty array)

SferaDev, line 867:
```suggestion
```

SferaDev: Added them back, seems there was a bit of delay yesterday.

Creating https://github.com/xataio/client-ts/issues/181 for follow-up

client/src/schema/repository.ts
SferaDev: I would like to offer both (inside and outside) as-is here.

Creating https://github.com/xataio/client-ts/issues/182 for improving this

SferaDev:
@gimenete Moving this again for review, I will follow-up on the improvements soon.

What do you think about the API in any case?

```ts
const users = await xata.db.users.search("Alberto")
const { users } = await xata.search("Alberto", { tables: "users" })
```

gimenete:
> What do you think about the API in any case?

I think that looks great 👍

gimenete (approved):
LGTM 👌

PR #167, opened 2022-05-11
Make init object helper private

SferaDev:


codegen/example/xata.ts
```diff
@@ -17,20 +23,23 @@ export interface User {
 
 export type UserRecord = User & XataRecord;
 
-const links = { teams: [['owner', 'users']], users: [['team', 'teams']] };
+const links = { teams: [["owner", "users"]], users: [["team", "teams"]] };
```
gimenete, line 26: 🤔 Why is this changing? We may want to run prettier in CI and see if the code is prettified. I can do that (in another PR).

gimenete (approved):

PR #173, opened 2022-05-11
Support env variables mode for codegen

gimenete:
With this PR the codegen package will use environment variables if they exist. Eventually this will be the only mechanism we support, but for backwards compatibility we want to keep using the schema and config files for a while.

The way it works is:

- If environment variables exist for the databaseURL and the api key, the schema is obtained from xata.io using the branch resolution algorithm.
- If not, it works as usual.

The package also loads the `.env` file in the current directory if it exists. And it's also possible to specify a different env file with `--env`. Example `--env env.local`.

TODO: use a stable version of `@xata.io/client`. This one probably: https://github.com/xataio/client-ts/pull/183


client/src/schema/index.test.ts
```diff
@@ -243,7 +246,7 @@ describe('request', () => {
       };
     });
 
-    await expect(users.getOne()).rejects.toThrowErrorMatchingInlineSnapshot(`"Not Found"`);
+    await expect(users.getOne()).rejects.toThrowErrorMatchingInlineSnapshot('"expected is not a function"');
```
SferaDev, line 249: My bad. Not for self to review this

codegen/package.json
```diff
@@ -25,11 +25,13 @@
   },
   "homepage": "https://github.com/xataio/client-ts/blob/main/codegen/README.md",
   "dependencies": {
+    "@xata.io/client": "0.0.0-alpha.3bb1ec0",
```
SferaDev, line 28: Probably a peerDependency here?

codegen/src/cli.ts
```diff
@@ -20,36 +21,23 @@ program
...
-      await mkdir(dir, { recursive: true });
-      await generateWithOutput({ xataDirectory, outputFilePath: out, spinner });
-    } catch (e) {
-      exitWithError(e);
+      const databaseURL = getDatabaseURL();
+      const apiKey = getAPIKey();
+      if (databaseURL && apiKey) {
+        await generateFromAPI(out, { databaseURL, apiKey });
+      } else {
+        await generateFromLocalFiles(xataDirectory, out);
```
SferaDev, line 36: Probably refactor to schemaFile (since we are removing the xataDirectory), still an offline copy of the online schema file could be used.

test/codegen.test.ts
```diff
@@ -1,28 +1,36 @@
...
+  it('should generate correct TypeScript', async () => {
+    const outputFilePath = 'test/example.ts';
+    await deleteFile(outputFilePath);
+    await generateFromLocalFiles(xataDirectory, outputFilePath);
 
-    const [path, content] = writeFile.mock.calls[0];
-    expect(path).toEqual(join(process.cwd(), 'hahaha.ts'));
-    expect(content).toMatchSnapshot();
+    expect(await readFile(outputFilePath, 'utf-8')).toMatchSnapshot();
+    await deleteFile(outputFilePath);
```
SferaDev, line 15: Note to self, this was a mock previously, made sense that way

client/src/schema/index.test.ts
fabien0102: Happens don't worry, this is why we have review 😉

SferaDev (approved):

codegen/package.json
gimenete: 🤔 Why do you prefer to install it as a `peerDependency`?

PR #184, opened 2022-05-12
Shell package

gimenete:
The shell opens a REPL with a `xata` global variable containing an initialized `XataClient` generated
for your current database and branch.

```shell
# The shell will use the env variables XATA_DATABASE_URL and XATA_API.
# You can define them in an `.env` file in the current directory.
npx @xata.io/shell
# Alternatively you can specify an env file
npx @xata.io/shell --env env.local
```

The shell uses the Node.js `repl` package with a couple of additions:

- Expressions that return a promise are resolved automatically.
- If you end an expression with `\t`, it will be printed with `console.table()`.

Example:

```js
// You can do
xata.db.teams.getMany() \t
// Which is equivalent to
console.table(await xata.db.teams.getMany())
```

TODO:
- use a stable version of codegen.
- In another PR I'll refactor the `scripts` we have to specify package versions to not duplicate them in each package.


client/src/schema/index.test.ts
```diff
@@ -243,7 +246,7 @@ describe('request', () => {
       };
     });
 
-    await expect(users.getOne()).rejects.toThrowErrorMatchingInlineSnapshot(`"Not Found"`);
+    await expect(users.getOne()).rejects.toThrowErrorMatchingInlineSnapshot('"expected is not a function"');
```
fabien0102, line 249: This diff looks suspicious ^^

codegen/src/codegen.ts
```diff
@@ -124,7 +123,7 @@ export async function generate({
       ${tables.map((table) => `"${table.name}": Repository<${getTypeName(table.name)}>;`).join('\n')}
     }> {
       constructor(options?: XataClientOptions) {
-        super({ databaseURL: "https://${config.workspaceID}.xata.sh/db/${config.dbName}", ...options}, links);
+        super({ databaseURL: "${databaseURL}", ...options}, links);
```
fabien0102, line 126:
```suggestion
        super({ databaseURL, ...options}, links);
```

package.json
```diff
@@ -4,8 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "jest",
-    "test:watch": "jest --watch",
+    "test": "vitest",
```
fabien0102, line 7: 👌

codegen/src/codegen.ts
SferaDev: Actually no, this is the codegen, so we need to inline the ``databaseURL``

package.json
SferaDev: The head branch was actually not correctly set, so the diff you were seeing was from https://github.com/xataio/client-ts/pull/173 😆

gimenete: Whops, my bad. Forgot to change the head branch when creating the PR.

SferaDev:
image

gimenete:
@SferaDev yeah, I'm aware. I'm trying to figure out why, because it is exported here https://github.com/xataio/client-ts/blob/main/codegen/src/index.ts#L2 and works locally, but not with npx 🤔

SferaDev (approved):

PR #190, opened 2022-05-13
Move common scripts to root dir

gimenete:
What the title says 😄


SferaDev (approved):
Awesome!

Now that we have many packages, I would like to propose a small folder refactor under a `packages` folder. With subfolder matching the npmjs name.

```
packages/client
packages/client-plugin-x
packages/codegen
packages/shell
packages/importer

packages/cli
packages/cli-plugin-x
```

Maybe we want to keep the CLI in a different repo, although I kind of like having it together as it's quite related and helps having shared infrastructure. 

WDYT @gimenete?

gimenete:
I like that idea! I'll create an issue for that 👍

PR #195, opened 2022-05-16
Use lerna bootstrap to make dependencies work with each other

gimenete:


SferaDev:
Found the solution with ``lerna publish --exact`` in another PR

PR #211, opened 2022-05-18
Export cjs and esm bundles

SferaDev:
Let's be nice with frontend developers in this chaotic times of ESM and CJS

### DIFF
--- a/packages/client/tsconfig.json
+++ b/packages/client/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES6",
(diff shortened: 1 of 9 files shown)


gimenete (approved):

packages/client/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES6",
+    "target": "ES2020",
```
SferaDev, line 3:
```suggestion
    "target": "ES6",
```

PR #214, opened 2022-05-18
Shell command

gimenete:
To test:

```
cd cli
npm run build
./bin/run.js shell
```

You need to have the environment variables set or have a .env file in the same dir you are running the command.


cli/src/commands/shell/index.ts
```diff
@@ -1,17 +1,10 @@
...
-  static flags = {};
-
-  static args = [];
+  static description = 'Open a shell to the current database and branch';
 
   async run(): Promise<void> {
-    // const { args, flags } = await this.parse(Shell);
-
-    this.error('To be done');
+    await run({});
```
SferaDev, line 8: Probably makes sense to make ``options`` optional

SferaDev (approved):

cli/src/commands/shell/index.ts
gimenete: Yep, yep. I'll do a PR for that.

PR #217, opened 2022-05-19
Fix search return type (XataRecord)

SferaDev:
- Enable search integration tests
- Return XataRecord initialized object for search results


SferaDev:
If we see that the search tests randomly fail we might need to disable them again

gimenete (approved):

PR #222, opened 2022-05-20
Add simple cache strategy + LRU plugin

SferaDev:
- Adds a new ``cache`` option to the ``Client`` that is later passed to any available plugin for usage
- Adds a new package ``@xata.io/client-plugin-cache`` with a ``LRUCache``. Ideally we will provide more caching strategies there too.

Improvements:
- ``Query.toString()`` right now is a ``JSON.stringify``, however we might want to build a 0-dependency hashing that is more predictable.


packages/plugin-cache/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@xata.io/client-plugin-cache",
+  "version": "0.0.1",
+  "private": true,
```
SferaDev, line 4:
```suggestion
```

We can commit this change as soon as we agree on the naming for the SDK plugins

packages/client/src/schema/repository.ts
```diff
@@ -182,24 +189,36 @@ export class RestRepository<Data extends BaseData, Record extends XataRecord = D
...
+
+      return records;
     }
 
     // Create one record with id as param
     if (isString(a) && isObject(b)) {
       if (a === '') throw new Error("The id can't be empty");
-      return this.#insertRecordWithId(a, b);
+      const record = await this.#insertRecordWithId(a, b);
+      await this.recordCache.set(record.id, record);
```
gimenete, line 202: Here and in any other place where we are storing single records, we need to generate a cache key that contains the table name and the record id, because the record id can be the same for records in different tables.

gimenete (commented):
Awesome work!! This is great.

There's one important thing to have into account for cache keys for single records. I left a comment.

Also, since caching can lead to reading / writing the wrong data and that's risky, it'd be great to have tests covering the caching functionality.

packages/plugin-cache/package.json
gimenete: What about `plugin-cache-memory`? The reason is:
- It'd be nice if all our plugins start with `plugin-`.
- And all the cache plugins start with `plugin-cache-` so it's easy to look for them alphabetically.
- I prefer "memory" because it makes it easy to understand that the caching is done in memory.

SferaDev: Will we have ``plugins`` for the CLI? I'd like to differentiate between client and cli plugins if that's the case.

I kind of like the ``plugin-cache-*``, although maybe ``plugin-cache-lru`` instead of ``plugin-cache-memory``.

gimenete: Good point about the CLI having plugins. Then `client-plugin-cache` or `plugin-client-cache` makes sense to me.

packages/client/src/schema/repository.ts
```diff
@@ -182,24 +189,36 @@ export class RestRepository<Data extends BaseData, Record extends XataRecord = D
   ): Promise<SelectedPick<Record, ['*']> | SelectedPick<Record, ['*']>[]> {
     // Create many records
     if (Array.isArray(a)) {
-      return this.#bulkInsertTableRecords(a);
+      const records = await this.#bulkInsertTableRecords(a);
+      await Promise.all(records.map((record) => this.recordCache.set(`${this.#table}-${record.id}`, record)));
```
gimenete, line 193: We still have potential clashing here, because `-` is an allowed character for tables.

So if you have table=foo-bar key=baz, and table=foo key=bar-baz, they clash.

Instead of `-` we can use `:` or something else.

gimenete (commented):
I've left a comment about clashing keys.

Also, one question. Are we taking into account that objects can have different set of columns selected?

SferaDev:
Renamed to ``plugin-client-cache`` and removed the ``private``flag, so it's published to npm.

Also fixed the key clashing with dashes to semicolons

SferaDev:
> Are we taking into account that objects can have different set of columns selected?

No, and yes 😃 

We don't have support on the CRUD methods for column selection, so no need. We might in the future but it's blocked due to https://github.com/xataio/openapi/issues/139.

In queries we support column selection, the key represents all the payload and that includes them.

packages/client/src/schema/repository.ts
```diff
@@ -452,9 +493,48 @@ export class RestRepository<Data extends BaseData, Record extends XataRecord = D
...
+    query: Query<Record, XataRecord>
+  ): Promise<{ meta: RecordsMetadata; records: T[] } | null> {
+    const key = `query_${this.#table}:${query.key()}`;
+    const result = await this.cache.get<{ date: Date; meta: RecordsMetadata; records: T[] }>(key);
+    if (!result) return null;
+
+    const { ttl } = query.getQueryOptions();
+    if (!ttl || ttl < 0) return result;
+
+    const hasExpired = result.date.getTime() + ttl < new Date().getTime();
```
gimenete, line 535: Nitpick

```suggestion
    const hasExpired = result.date.getTime() + ttl < Date.now();
```

gimenete (commented):
If I understand correctly from the code, right now with this implementation if you set a cache, all record operations are cached or none are cached. There's no way to opt-in, opt-out individually by operation, right?

Sorry if this is a big change but I think we should only implement caching in queries (not for single records) and if it's specified explicitely when building the query. Example:

```ts
xata.db.teams.cache(ttl).filter(...).getMany()
```

I think it'd be also be useful to know if the result is coming from the cache or not. For that we could add another argument to `cache()`.

```ts
const query = xata.db.teams.cache(ttl).filter(...)
const cached = await query.getCached()
const fresh = await query.cache(0).getMany()
```

The problem with single records is that it can easily produce errors due to not reading fresh copies of the data. Besides, usually querying a single record is fast. Typically you just want to cache queries because they may be expensive.

WDYT?

packages/client/src/schema/query.ts
```diff
@@ -68,6 +70,12 @@ export class Query<Record extends XataRecord, Result extends XataRecord = Record
     return this.#data;
   }
 
+  key(): string {
+    const { columns = [], filter = {}, sort = [], page = {} } = this.#data;
+    const key = JSON.stringify({ columns, filter, sort, page });
+    return key;
```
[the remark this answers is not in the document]
SferaDev: I actually have a helper for that, I removed it while debugging.

SferaDev:
Updated the default behaviour, let me know what do you think.

Default: CRUD operations are cached, Queries are not cached

Queries can be cached with this API:

```
xata.db.teams.getMany({ ttl: 2 * 60 * 1000 })
```

SferaDev:
Okay so finally:

Default behaviour:

- Nothing is cached
- Queries can opt in to be cached with ``getX({ cache: numberInMs })`` or ``.cache(numberInMs).getX()``

Examples:

```
xata.db.teams.cache(ttl).filter(...).getMany()
xata.db.teams.filter(...).getMany({ cache: ttl })
```

> I think it'd be also be useful to know if the result is coming from the cache or not.

I wouldn't personally add a ``query.getCached()`` method. If you really need to add inspect the cache, you can always create your own and pass it to the client in its constructor. I don't believe anyone needs to do it, but they are capable:

```
const cache = new SimpleCache();
const xata = new XataClient({ cache });
cache.has/cache.get/cache.delete
```

> The problem with single records is that it can easily produce errors due to not reading fresh copies of the data. Besides, usually querying a single record is fast. Typically you just want to cache queries because they may be expensive.

`CacheImpl` now has a property called `cacheRecords` to enable/disable single record caching. Example

```
const xata = new XataClient(); // Record cache disabled. Default TTL for queries is 0. 
const xata = new XataClient({ cache: new SimpleCache() }); // Record cache enabled. Default TTL for queries is 1 minutes.
const xata = new XataClient({ cache: new SimpleCache({ defaultQueryTTL: 10 * 1000, cacheRecords: false }) }) // Record cache configurable. Default TTL for queries configurable.
```

The @xata.io/plugin-client-cache will offer other caching libraries. Such as ``lru-cache`` that depend on other libraries (marked as peerDependencies).

gimenete:
> I wouldn't personally add a query.getCached() method. If you really need to add inspect the cache, you can always create your own and pass it to the client in its constructor. I don't believe anyone needs to do it

I am thinking on a use case like this: you have a Next.js application and you want to return a query result in a `getServerSideProps` but only if it's cached to make the time to render as fast as possible. Then via other mechanism (API route) the client calls the backend to get the most up to date data by using the same or equivalent Query object. In this case without caching.

gimenete (approved):
Great work!!

I've left a comment on getting the cached result only. And also, I wouldn't do a default ttl, but always make the caching as an opt-in. But both things are non blockers. We can iterate on them.

SferaDev:
> I've left a comment on getting the cached result only.

It's not easy to get the cached result only in the query level, if we find it useful in the future we can add it.

> I wouldn't do a default ttl, but always make the caching as an opt-in.

Right now it does not have a default ttl, it's opt-in only.

If the user explicitly enables a cache then, and only then, it has a TTL (that they can optionally configure)

PR #224, opened 2022-05-20
CLI. Implement a few list commands

gimenete:
```
./bin/run.js workspaces list
./bin/run.js branches list
./bin/run.js dbs list
```


cli/src/base.ts
```diff
@@ -0,0 +1,38 @@
...
+import ansiRegex from 'ansi-regex';
+import chalk from 'chalk';
+import dotenv from 'dotenv';
+import fetch from 'node-fetch';
+import table from 'text-table';
+import { readAPIKey } from './key.js';
+
+export abstract class BaseCommand extends Command {
+  locale: string | undefined = undefined;
+  timeZone: string | undefined = undefined;
```
SferaDev, line 12: Can we somehow init this from system env variables?

cli/src/base.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  async init() {
+    dotenv.config();
+  }
+
+  async getXataClient() {
+    const apiKey = await readAPIKey();
+    if (!apiKey) this.error('Could not instantiate Xata client. No API key found.'); // TODO: give suggested next steps
```
SferaDev, line 20: I know we partially discussed it last week, and I saw your points on using env variables for API Keys.

BUT... After a second thought, I really think that it would improve the DX of the CLI that if there's no env variable, we prompt the user with a URL to the frontend, that grants a new API Key and the CLI starts a server to receive it as in OAuth. Finally it stores for re-use in an encrypted config.

WDYT? We can probably discuss this in a meeting, and no actions needed in this PR!

cli/src/commands/branches/list.ts
```diff
@@ -1,15 +1,42 @@
...
+    })
+  };
 
   static args = [];
 
-  async run(): Promise<void> {
-    this.error('To be done');
+  static enableJsonFlag = true;
+
+  async run(): Promise<any> {
```
SferaDev, line 23: Not 100% about having ``any`` with the command output (applies to the rest of commands too.

I think we should have something as ``Branch[] | undefined``. We could even add a helper type ``Maybe<Branch[]>``.

If we won't never call the commands classes directly outside of testing (I think some commands will benefit from composition), maybe we can just remove the return type and let TS infer it.

cli/src/defaults.ts
```diff
@@ -0,0 +1,13 @@
+import { getDatabaseURL } from '@xata.io/client';
```
SferaDev, line 1: [really-minor] Not sold on the name ``defaults.ts``, but no better suggestions either.

cli/src/base.ts
SferaDev: Edit: Just saw https://github.com/xataio/client-ts/pull/224/files#diff-6a5236bbaf544aebc3a41c104350ad36d5c5a779eeebc40da558de9bbf893f0bR10

Okay, all in now. I think it should still be encrypted and we should offer some managed way to set it up with the frontend but that's the future.

cli/src/base.ts
gimenete: This is for date formatting. Passing `undefined` will use the system's default.

I put this only for snapshosts in tests, to have consistency between runs I'm setting the locale and timezone to the same values just while testing.

I'll leave a comment to explain this in the code.

SferaDev (approved):
We're slowly getting there!

I think we should have an interactive way to traverse the list table with inquiries. Like enter a workspace, enter a database, enter a branch, and have actions all the way there. But that definitely is not a list command and not a requirement right now.

Maybe it would be nice to print some info of the path where the information is being retrieved. For example, if we do a ``branches list`` that it shows the workspace/database first before printing the table.

cli/src/base.ts
gimenete: Yeah, I'm using the same path than the go xata cli. This is pretty standard: storing a key in a file with read-only permissions for the current user. See https://github.com/xataio/cli/blob/main/config/config.go#L23

I think it's also a good idea to try to read an env variable, specially if you have an `.env` file, to use a key that is scoped to your database potentially and not having to have a broad API key.

I'm open to discussing all this before releasing the first version 👍

cli/src/commands/branches/list.ts
gimenete: Sounds good 👍

gimenete: I think I'll do it in another PR.

Issue #229, opened 2022-05-22
Shell does not have type-safety


TejasQ:
Sorry if this is a known issue, please feel free to close if it is.

Repro:

https://user-images.githubusercontent.com/9947422/169699254-4a427b9a-95ff-4162-a0b4-665ed52ca4f7.mov

SferaDev:
Yeah, REPL is not type safe in any case (not TypeScript safe). Although the proxy hides the namespaces.

We briefly discussed this with @gimenete last Thursday. And to improve the DX we can either tweak a little bit the Proxy target or play with the suggestions in the REPL.

Good to have the issue tracking though

PR #230, opened 2022-05-23
Allow introspecting tables in shell

SferaDev:
Closes: #229

image
image
image

### DIFF
--- a/packages/codegen/src/codegen.ts
+++ b/packages/codegen/src/codegen.ts
@@ -104,6 +104,8 @@ export async function generate({
 
     const links = ${JSON.stringify(links)};
 
+    const tables = [${tables.map((table) => `'${table.name}'`).join(', ')}];
+
     ${language === 'javascript' ? `/** @type { import('@xata.io/client').ClientConstructor<{}> } */` : ''}
     const DatabaseClient = buildClient();
 
@@ -115,7 +117,7 @@ export async function generate({
     }
     export class XataClient extends DatabaseClient<DatabaseSchema> {
       constructor(options?: BaseClientOptions) {
-        super({ databaseURL: "${databaseURL}", ...options}, links);
+        super({ databaseURL: "${databaseURL}", ...options}, links, tables);
       }
     }
   `;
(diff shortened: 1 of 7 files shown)


packages/codegen/src/codegen.ts
```diff
@@ -115,7 +117,7 @@ export async function generate({
     }
     export class XataClient extends DatabaseClient<DatabaseSchema> {
       constructor(options?: BaseClientOptions) {
-        super({ databaseURL: "${databaseURL}", ...options}, links);
+        super({ databaseURL: "${databaseURL}", ...options}, links, tables);
```
SferaDev, line 120: To be honest, I would either switch order here or make it an object. Haven't done it to not break compatibility with existing code-generated clients, but maybe it's worth. WDYT?

packages/shell/src/cli-run.ts
```diff
@@ -43,6 +43,7 @@ export async function run(options: { env?: string; databaseURL?: string; apiKey?
   const defaultEval = getDefaultEval();
   const replServer = repl.start({
     preview: true,
+    ignoreUndefined: true,
```
SferaDev, line 46: Including also this, unrelated to the main PR goal

gimenete: 🤔 I played with that option and I preferred to leave it disabled. The reason is that when you press enter multiple times you just see a list of prompts (`>`) and it's confusing because you don't know if there's a command waiting to be closed (a missing `)` or `}`, etc.) or it's just that there's nothing to execute.

packages/codegen/src/codegen.ts
gimenete: I'd leave it like it is for now. It'd make more sense to change the order but it's not a big deal and personally I don't think it's worth it at the moment because we need to deal with backwards compatibliity, like you say.

gimenete (approved):
LGTM. I'd prefer to not enable `ignoreUndefined` but I have no strong opinions.

SferaDev:
> I'd prefer to not enable ignoreUndefined but I have no strong opinions.

No strong opinions here either, felt more natural without but you have a point with knowing if the parser believes input is complete

PR #232, opened 2022-05-23
CLI: delete commands

gimenete:
```
./bin/run.js workspaces delete
./bin/run.js branches delete
./bin/run.js dbs delete
```


cli/src/commands/dbs/delete.ts
```diff
@@ -1,15 +1,51 @@
...
+    })
+  };
 
   static args = [];
 
-  async run(): Promise<void> {
-    this.error('To be done');
+  static enableJsonFlag = true;
+
+  async run(): Promise<void | unknown> {
```
gimenete, line 24: I'll change these types all in one PR.

cli/src/commands/branches/delete.ts
```diff
@@ -1,15 +1,67 @@
...
+    const workspace = flags.workspace || defaults.workspace;
+    const database = flags.database || defaults.database;
+    const branch = flags.branch || (await this.getBranchName(databaseURL));
+
+    if (!workspace)
+      return this.error('Could not find workspace id. Please set XATA_DATABASE_URL or use the --workspace flag.');
+    if (!database)
+      return this.error('Could not find database name. Please set XATA_DATABASE_URL or use the --database flag.');
+    if (!branch)
+      return this.error('Could not find branch name. Either set up a git repository or use the --branch flag.');
```
SferaDev, line 41: I wonder if we could make this reusable in the BaseCommand so we don't need the check on every command.

Something like:

const { workspace, database } = this.parseContext(flags, ["workspace", "database"]);

cli/src/commands/branches/delete.ts
```diff
@@ -1,15 +1,67 @@
...
+
+    const xata = await this.getXataClient();
+
+    const result = await prompts({
+      type: 'confirm',
+      name: 'confirm',
+      message: `Are you sure you want to delete the branch ${database}:${branch} in the ${workspace} workspace?`,
+      initial: true
+    });
+    if (!result.confirm) return this.exit(1);
```
SferaDev, line 51: Is it an error if the user cancels? Not sure about best practices. Also very minor as, this only happens interactively.

cli/src/commands/branches/delete.ts
```diff
@@ -1,15 +1,67 @@
...
+    if (!workspace)
+      return this.error('Could not find workspace id. Please set XATA_DATABASE_URL or use the --workspace flag.');
+    if (!database)
+      return this.error('Could not find database name. Please set XATA_DATABASE_URL or use the --database flag.');
+    if (!branch)
+      return this.error('Could not find branch name. Either set up a git repository or use the --branch flag.');
+
+    const xata = await this.getXataClient();
+
+    const result = await prompts({
```
SferaDev, line 45: Does this prompts library support the bypass flag (--yes --non-interactive --ci) that automatically returns the `initial` option? Should we wrap it somehow and use our own this.prompt()?

cli/src/commands/workspaces/delete.ts
```diff
@@ -1,15 +1,45 @@
...
+      type: 'confirm',
+      name: 'confirm',
+      message: `Are you sure you want to delete workspace ${workspace}?`,
+      initial: true
+    });
+    if (!result.confirm) return this.exit(1);
+
+    await xata.workspaces.deleteWorkspace(workspace);
+
+    if (this.jsonEnabled()) return {};
```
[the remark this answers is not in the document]
SferaDev: Also, should we return the result of the api call? In case that we end up returning something? Also it's easier to type

SferaDev (approved):
LGTM, some comments nothing that cannot be done in the future

cli/src/commands/branches/delete.ts
gimenete: I can't find an example now, but yeah, most CLIs I think exit with an error code if the user cancels.

cli/src/commands/branches/delete.ts
gimenete: Great question. `prompts` doesn't do that, but we'll definitely need to add a `--no-input` option. I'll see how to do that. Probably with utility methods, yep (one for ask for input, another to ask for confirmation probably). I want to do this in another PR once all/most commands are implemented with interactivity to see all the needs together.

cli/src/commands/workspaces/delete.ts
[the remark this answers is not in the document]
gimenete: > I imagine we will add the json flag everywhere

Not necessarely. I expect some commands to be always interactive and not have a --json flag. For example editing the schema, or authentication.

Maybe we can have a class that extends BaseCommand for all those commands. I'll add this to the list of improvements.

> should we return the result of the api call?

Mostly, yes. I'm not sure about some commands. For example `ListDatabasesResponse` has a nested `databases` field only. Not sure if it's worth printing the whole response, or just the `databases` part.

gimenete:
Thanks for the review @SferaDev !!

I've answered all the comments and I've created https://github.com/xataio/client-ts/issues/237 to track some of the feedback.

PR #233, opened 2022-05-23
CLI: create commands

gimenete:
```
./bin/run.js workspaces create
./bin/run.js branches create
./bin/run.js dbs create
```


cli/src/commands/workspaces/create.ts
```diff
@@ -7,9 +8,24 @@ export default class WorkspacesCreate extends BaseCommand {
...
+    const { args } = await this.parse(WorkspacesCreate);
+    const { workspace } = args;
+
+    if (!workspace) {
+      return this.error('Please, specify a workspace name');
+    }
+
+    const xata = await this.getXataClient();
+
+    const result = await xata.workspaces.createWorkspace({ name: workspace, slug: slugify(workspace) });
```
SferaDev, line 25: Do we want to allow the user to send through flags more parameters like slug? Even if we default to the slugify

SferaDev (approved):
The code looks good. Haven't fully tested the git integration (and it's edge cases), in any case approving as we can improve it later on

cli/src/commands/workspaces/create.ts
gimenete: Good suggestion!

Another option is that the backend makes the `slug` field optional and is calculated in the server side. But anyway, maybe it's a good idea to allow users to specify it. I'll add to the _improvements_ issue.

PR #236, opened 2022-05-24
Add API support to the shell

SferaDev:
Add API methods to the shell:

```
api.workspaces.createWorkspace({ name: "foo", slug: "bar" })
```

```
GET /dbs
```


packages/shell/src/cli-run.ts
```diff
@@ -117,3 +147,15 @@ async function setupHistory(replServer: repl.REPLServer) {
...
+  databaseURL = databaseURL || getDatabaseURL() || '';
+  const [protocol, , host, , database] = databaseURL.split('/');
+  const [workspace] = (host || '').split('.');
+  return {
+    protocol,
+    host,
+    database,
+    workspace
+  };
+}
```
SferaDev, line 161: This is from the CLI, I think more and more a util package would be appreciated 😄

gimenete: Yep. Or we can put it in the `xata/client` one.

packages/shell/src/cli-run.ts
```diff
@@ -40,6 +41,27 @@ export async function run(options: { env?: string; databaseURL?: string; apiKey?
     )} global variable.`
   );
 
+  const fetchApi = async (method: string, path: string, body?: any) => {
+    const { protocol, host } = parseDatabaseURL(databaseURL);
+    // TODO: Add support for staging, how?
```
gimenete, line 46: What about getting the `host` returned by `parseDatabaseURL`?

gimenete (approved):
This is great! 😍 

I was thinking on maybe having a different command/REPL, but I think it is not necessary now. It's pretty clear what is an API call and what is a JS expression, and it's cool to have both combined and be able to use both at the same time.

Great job!! 🚀

packages/shell/src/cli-run.ts
SferaDev: Yeah, that could work. The CLI uses an env variable for changing the API host. Maybe that's a better approach. I'll leave the comment and let's revisit it with usage.

PR #238, opened 2022-05-24
Update dependencies workflow

SferaDev:
- Add ``--force`` flag to ``npm install``, needed when peer dependencies fail
- Add ``--deep``flag to ``ncu`` to streamline with ``client-ts``
- Use the same action to create PRs as in other workflows
- Add a description with the update to the body


gimenete (approved):
🚀

PR #242, opened 2022-05-25
Fix npm run build-example

gimenete:
There were files being ignored and files not in the right place.


gimenete:
The way we build the example will change here https://github.com/xataio/client-ts/pull/341

Issue #237, opened 2022-05-24
CLI improvements


gimenete:
- [ ] Base class for all commands that support `--json`.
- [ ] Better typing for `run()` methods.
- [x] Utility methods for `prompts` and skip interactivity if `--no-input` is used or the terminal is not a TTY.
  - [ ] https://www.npmjs.com/package/@npmcli/ci-detect
- [ ] Always return the full API response?
- [x] Utilities for parsing flags that have also defaults depending on the context: `workspace`, `database`, `branch`.
- [ ] Allow users to specify slug when creating a db.
- [ ] Add a _verbose_ option to print things such as the config file being used.
- [ ] Document exit codes.
- [ ] Validate identifiers before sending them to the server?
- [ ] [Support pnpm when installing the SDK in `xata init`](https://github.com/xataio/client-ts/issues/643)
- [x] Check if there's an `API_KEY` already in the `.env` file when writing to it
- [ ] Support other paths for the `.env` file
- [ ] Offer a command to generate/update API Key in the .env file without having to run `xata init`
- [x] Add examples.
- [x] Use suggestions in errors.

SferaDev:
- [ ] Save database url in project config file
- [ ] Instead of failing if no database url found, inquiry user with his available workspaces and databases

PR #243, opened 2022-05-25
CLI: codegen

gimenete:
```
./bin/run.js codegen src/xata.ts
./bin/run.js codegen # Uses cosmiconfig
```

### DIFF
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -5,7 +5,19 @@ import chalk from 'chalk';
 import dotenv from 'dotenv';
 import fetch from 'node-fetch';
 import table from 'text-table';
+import { z } from 'zod';
 import { readAPIKey } from './key.js';
+import { cosmiconfigSync } from 'cosmiconfig';
+
+export const userConfig = z.object({
+  codegen: z.optional(
+    z.object({
+      output: z.string()
+    })
+  )
+});
+
+export type UserConfig = z.infer<typeof userConfig>;
 
 export abstract class BaseCommand extends Command {
   // Date formatting is not consistent across locales and timezones, so we need to set the locale and timezone for unit tests.
@@ -13,8 +25,24 @@ export abstract class BaseCommand extends Command {
   locale: string | undefined = undefined;
   timeZone: string | undefined = undefined;
 
+  userConfig: UserConfig | undefined;
+
   async init() {
     dotenv.config();
+
+    const search = cosmiconfigSync('xata').search();
+    if (search) {
+      const result = userConfig.safeParse(search.config);
+      if (result.success) {
+        this.userConfig = result.data;
+      } else {
+        this.warn(`The configuration file ${search.filepath} was found, but could not be parsed:`);
+
+        for (const error of result.error.errors) {
+          this.warn(`  [${error.code}] ${error.message} at "${error.path.join('.')}"`);
+        }
+      }
+    }
   }
 
   async getXataClient() {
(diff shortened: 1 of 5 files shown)


cli/src/base.ts
```diff
@@ -5,16 +5,44 @@ import chalk from 'chalk';
...
   locale: string | undefined = undefined;
   timeZone: string | undefined = undefined;
 
+  userConfig: UserConfig | undefined;
+
   async init() {
     dotenv.config();
+
+    const search = cosmiconfigSync('xata').search();
+    if (search) {
```
gimenete, line 34: Maybe we can add a _verbose_ option that prints the filepath of the config.

cli/src/base.ts
```diff
@@ -5,16 +5,44 @@ import chalk from 'chalk';
...
+      const result = userConfig.safeParse(search.config);
+      if (result.success) {
+        this.userConfig = result.data;
+      } else {
+        this.warn(`The configuration file ${search.filepath} was found, but could not be parsed:`);
```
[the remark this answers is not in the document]
gimenete: Added an item here https://github.com/xataio/client-ts/issues/237

cli/src/base.ts
gimenete, line 34: Added an item here https://github.com/xataio/client-ts/issues/237 to print the filepath found with a _verbose_ option.

cli/src/base.ts
```diff
@@ -5,16 +5,44 @@ import chalk from 'chalk';
 import dotenv from 'dotenv';
 import fetch from 'node-fetch';
 import table from 'text-table';
+import { z } from 'zod';
 import { readAPIKey } from './key.js';
+import { cosmiconfigSync } from 'cosmiconfig';
+
+export const userConfig = z.object({
```
SferaDev, line 12: Probably a better name is projectConfig

cli/src/commands/codegen/index.ts
```diff
@@ -0,0 +1,54 @@
+import { generateFromContext } from '@xata.io/codegen';
+import { mkdir, writeFile } from 'fs/promises';
+import fetch from 'node-fetch';
+import path, { dirname, extname, relative } from 'path';
+import { BaseCommand } from '../../base.js';
+
+const langauges: Record<string, 'javascript' | 'typescript'> = {
```
SferaDev, line 7: Typo in languages

cli/src/commands/codegen/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  static examples = [];
+
+  static flags = {};
+
+  static args = [{ name: 'output', description: 'The output file to generate', required: false }];
+
+  async run(): Promise<void> {
+    const { args } = await this.parse(Codegen);
+
+    const output = args.output || this.userConfig?.codegen?.output;
```
SferaDev, line 25: [improvement for later]


If output is set by args, add an inquiry to set/update the config file.

If no config file yet, inquiry to initialize it with possible supported file names: .xatarc, xata.config.js....

SferaDev (approved):

cli/src/base.ts
gimenete: Great suggestion! I'll change that.

cli/src/commands/codegen/index.ts
gimenete: Yep. I've been thinking about this. I'd also like to add `declarations: true/false` and maybe in the near future plugins? One idea is to not allow arguments at all 🤔 with so many options and because if we always rely on the configuration file we have an ideal scenario to call the codegen when the schema changes, the user switches branches, etc.

I'm also thinking on adding `xata config get/set`. Similar to git config. Not sure if comsiconfig offers something like that though. I'll investigate.

But yeah, we need to provide a mechanism with prompts to configure those options if they are not set.

PR #245, opened 2022-05-25
CLI: auth

gimenete:
```
./bin/run.js auth login
./bin/run.js auth logout
./bin/run.js auth status
```

TODO: tests

### DIFF
--- a/cli/src/commands/auth/login.ts
+++ b/cli/src/commands/auth/login.ts
@@ -32,13 +32,7 @@ export default class Login extends BaseCommand {
     });
     if (!result.key) this.exit(2);
 
-    this.log('Checking access to the API...');
-    const xata = await this.getXataClient(result.key);
-    try {
-      await xata.workspaces.getWorkspacesList();
-    } catch (err) {
-      return this.error(`Error accessing the API: ${err instanceof Error ? err.message : String(err)}`);
-    }
+    await this.verifyAPIKey(result.key);
 
     await writeAPIKey(result.key);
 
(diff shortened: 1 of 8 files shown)


cli/src/commands/auth/login.ts
```diff
@@ -0,0 +1,41 @@
+import prompts from 'prompts';
+import { BaseCommand } from '../../base.js';
+import { readAPIKeyFromFile, writeAPIKey } from '../../key.js';
+
+export default class Login extends BaseCommand {
+  static description = 'Authenticate with Xata';
+
+  static examples = [];
+
+  static flags = {};
```
SferaDev, line 10: Add a flag to login without interactivity

gimenete: That'd be handy, but my concern is that the API key will be stored in the command history in the shell.

cli/src/commands/auth/login.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  async run(): Promise<void> {
+    const existingKey = await readAPIKeyFromFile();
+    if (existingKey) {
+      const result = await prompts({
+        type: 'confirm',
+        name: 'overwrite',
+        message: 'Authentication is already configured, do you want to overwrite it?'
+      });
+      if (!result.overwrite) this.exit(2);
```
SferaDev, line 22: Are we documenting error codes somewhere?

SferaDev (approved):

cli/src/commands/auth/login.ts
gimenete: Good point. I'll add a note on the issue for improvements.

PR #248, opened 2022-05-26
CLI: init, config set/get and a few clean ups

gimenete:
This PR implmenets the `xata init` command, `xata config get/set`, and a few clean ups.

This PR also changes the way workspace and database flags work for other commands. Now, commands that require a workspace AND database and the database is not the target of the action, can receive a databaseURL flag, or the databaseURL value from the project config is used (this is also new).

Besides, if no databaseURL, workspace or database can be calculated the user is prompt to interactively select them.

This PR also adds some changes regarding flags:
- `--json`, manually, becuse the built-in one it was failing in commands that didn't have an empty flags array. This looks like a bug in oclif. Maybe we could remove the built-in mechanism entirely in another PR or try to fix it in oclif.
- `--no-input`. Still not used, but in another PR will avoid interactivity.


cli/package.json
```diff
@@ -31,15 +31,21 @@
     "chalk": "^5.0.1",
     "cosmiconfig": "^7.0.1",
     "dotenv": "^16.0.1",
+    "lodash.get": "^4.4.2",
+    "lodash.set": "^4.3.2",
```
gimenete, line 35: This is for implementing `xata conifg get/set`. Not sure if I'll do it in this PR. I'll remove this if I don't.

cli/src/base.ts
```diff
@@ -75,4 +98,118 @@ export abstract class BaseCommand extends Command {
...
+      }));
+
+      if (options.allowCreate) {
+        choices.splice(0, 0, { title: 'Create a new database', value: 'create' });
+      }
+
+      const result = await prompts({
+        type: 'select',
+        name: 'database',
+        message: dbs.length > 0 && options.allowCreate ? 'Select a database or create a new one' : 'Select a database',
```
SferaDev, line 156: Same as before, instead of an allowCreate flag. Show the N databases + 1 option to create a new one

cli/src/commands/init/index.ts
```diff
@@ -1,17 +1,163 @@
...
+        name: 'output',
+        message: 'Choose where the output file for the code generator',
+        initial: 'src/xata.ts'
+      },
+      {
+        type: 'confirm',
+        name: 'declarations',
+        message: 'Do you want to generate the TypeScript declarations?',
+        initial: (prev) => !prev.endsWith('.ts')
+      }
```
SferaDev, line 84: Is there a way to conditionally hide questions? If it's TypeScript there's no need for a declarations file. This second question looks redundant to me

cli/src/commands/init/index.ts
```diff
@@ -1,17 +1,163 @@
...
+        `Looks like ${packageManager} is not installed or is not in the PATH. This made impossible to install the code generator`
+      );
+    }
+    this.packageManager = packageManager;
+    return packageManager;
+  }
+
+  async access(path: string) {
+    try {
+      await access(path);
```
SferaDev, line 114: What happens if I call xata from a subdir in my project? This only checks on the current folder, right?

SferaDev (commented):
Added some comments, nothing an error per-se, just challenging a bit the implementation and raising topics for discussion 😉

cli/src/base.ts
```diff
@@ -75,4 +98,118 @@ export abstract class BaseCommand extends Command {
...
+      message: 'Select a workspace',
+      choices: workspaces.workspaces.map((workspace) => ({
+        title: workspace.name,
+        description: workspace.id,
+        value: workspace.id
+      }))
+    });
+
+    return String(result.workspace);
+  }
```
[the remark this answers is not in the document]
gimenete: The behavior for workspaces is copied from here https://github.com/xataio/cli/blob/main/cmd/init.go#L57

I guess typically people would have one workspace, because when you sign up we are going to create a workspace for you. And unless you want to collaborate somehow with other people you won't create more workspaces.

Happy to discuss this on Monday during our sync if you want.

cli/src/commands/init/index.ts
gimenete: In this case you are initalizing the project, so it is supposed you want to initialize it in the current dir. Like when you do `npm init`, it creates a `package.json` file in the current dir.

cli/src/commands/dbs/list.ts
```diff
@@ -1,13 +1,13 @@
...
 import { BaseCommand } from '../../base.js';
-import { parseDatabaseURL } from '../../defaults.js';
 
 export default class DatabasesList extends BaseCommand {
   static description = 'List your databases';
 
   static examples = [];
 
   static flags = {
+    json: Flags.boolean({}),
```
SferaDev, line 10: Does this make sense?

```suggestion
    ...this.commonFlags,
```

cli/src/commands/init/index.ts
```diff
@@ -1,17 +1,292 @@
...
+import EditSchema from '../schema/edit.js';
 
 export default class Init extends BaseCommand {
-  static description = 'Initialize a database';
+  static description = 'Configure your working directory to work with a Xata database';
 
   static examples = [];
 
-  static flags = {};
+  static flags = {
```
SferaDev, line 20: I'd love init to also be non-interactive, later improvement of course

cli/src/commands/init/index.ts
```diff
@@ -1,17 +1,292 @@
...
+      message: 'Do you want to use the TypeScript/JavaScript code generator?',
+      initial: true
+    });
+    if (!confirm) return;
+
+    this.projectConfig = this.projectConfig || {};
+    this.projectConfig.codegen = {};
+
+    const { output } = await prompts({
+      type: 'text',
```
SferaDev, line 98: [idea] Would be cool that this allowed you to navigate existing folders, select existing file or create new file interactively instead of text path.

cli/src/commands/init/index.ts
```diff
@@ -1,17 +1,292 @@
...
+    const apiKey = await readAPIKey();
+
+    let content = '';
+    try {
+      // TODO: check if the file already contains the key and warn if it does
+      content = await readFile('.env', 'utf-8');
+    } catch (err) {
+      // ignore
+    }
+    content += `\n\nXATA_API_KEY=${apiKey}`;
```
SferaDev, line 183: We could probably ``detectNewlineAtEof`` and conditionally add 1 or 2 newlines. Maybe a comment too?

What if an API_KEY is already defined in the .env file?

cli/src/commands/init/index.ts
```diff
@@ -1,17 +1,292 @@
...
+  }
+
+  async writeEnvFile() {
+    // TODO: generate a database-scoped API key
+    const apiKey = await readAPIKey();
+
+    let content = '';
+    try {
+      // TODO: check if the file already contains the key and warn if it does
+      content = await readFile('.env', 'utf-8');
```
SferaDev, line 179: We should probably add a flag to customize the ``.env`` file name, and probably we could even detect all ``.env.*`` files and list them if there's more than one already in the project.

cli/src/commands/init/index.ts
```diff
@@ -1,17 +1,292 @@
...
+        choices: this.searchPlaces.map((place) => ({ title: place, value: place }))
+      });
+      if (!location) return this.error('You must select a location for the configuration file');
+
+      this.projectConfigLocation = path.join(process.cwd(), location);
+    }
+    await this.updateConfig();
+  }
+
+  async writeEnvFile() {
```
SferaDev, line 172: We could offer a command to generate/update API Key in the ``.env`` file but that doesn't require you to run ``init``

cli/src/schema.ts
```diff
@@ -0,0 +1,46 @@
...
+export type Column = {
+  name: string;
+  type: 'bool' | 'int' | 'float' | 'string' | 'text' | 'email' | 'multiple' | 'link' | 'object';
+  unique?: boolean;
+  description?: string;
+  link?: {
+    table: string;
+  };
+  columns?: Column[];
+};
```
SferaDev, line 13: We should probably use the types from ``@xata.io/client``, so that tsc errors out when we add new types (like date soon)

SferaDev (approved):
LGTM. Haven't found any blocking comments

cli/src/commands/init/index.ts
gimenete: Ohhh, nice idea. That'd be super cool. I'll see if there's a plugin for that, or we could try creating ours.

cli/src/base.ts
```diff
@@ -1,54 +1,74 @@
...
+export const projectConfigSchema = z.object({
+  databaseURL: z.string(),
+  codegen: z.object({
+    output: z.string(),
+    declarations: z.boolean()
+  })
 });
 
-export type ProjectConfig = z.infer<typeof projectConfig>;
+const partialProjectConfig = projectConfigSchema.deepPartial();
```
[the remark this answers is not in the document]
gimenete: It just changes where the deepPartial() is, right? The reason to put it there is because I want the schema without the optionals for the `config set` command. With the non-optionals schema I can easily traverse the schema to get the type of a specific config setting. With the partial schema it was more complicated and verbose. So basically I have both versions to different use cases.

cli/src/commands/dbs/list.ts
gimenete: Yep. I was testing with this file and forgot to update it 😅

cli/src/commands/init/index.ts
```diff
@@ -1,17 +1,292 @@
...
+      });
+
+      if (declarations) {
+        this.projectConfig.codegen.declarations = true;
+      }
+    }
+  }
+
+  async getPackageManager() {
+    const packageManager = (await this.access('yarn.lock')) ? 'yarn' : 'npm';
```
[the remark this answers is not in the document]
gimenete: I'll add it to the improvements issue. I have zero experience with pnpm. Is there a file we can use to detect its usage? Or we would just check if the binary exists in the PATH?

cli/src/commands/init/index.ts
gimenete: 👍  I'm putting these things about the .env file in the improvements issue.

cli/src/schema.ts
gimenete: Yeah, I'm not sure how to do that and use it in the zod schema.

I've tried this but I get a weird error

image

In any case, if the types don't match we will get a compilation error because we use the zod-generated schema when calling `xata.branches.getBranchMigrationPlan`. If the API spec schema and the zod schema are not compatible, we will know. However it'd be ideal to have the types from the API spec directly, I agree.

PR #259, opened 2022-06-03
CLI: Interactive schema editing

gimenete:
```
./bin/run.js schema edit
```

This command will open an interactive schema editor.

If there's a project file it will also run codegen after the schema is updated.

https://user-images.githubusercontent.com/50486/171819848-b1f3a723-dead-4225-9401-0b91e6f8f3f4.mp4


SferaDev (approved):
Awesome work!

When creating a column it's taking undefined if you don't enter a name:

Screenshot 2022-06-07 at 12 25 58

It failed running the migration though, but would be nice to not allow the user to leave the page with an bad state.

Screenshot 2022-06-07 at 12 27 07

Other than that it worked really great

gimenete:
@SferaDev thanks for the review! I've fixed the undefined column names here https://github.com/xataio/client-ts/pull/259/commits/9a200b32448ed4dc068037ba229bff1d736f601a

It's weird that `/^[a-zA-Z0-9-_~]+$/.test(undefined)` returns `true` 😱

PR #261, opened 2022-06-06
Fix CloudFlare workers require

gimenete:
Avoid `Detected a Node builtin module import while Node compatibility is disabled` in CloudFlare workers.

Introduced here https://github.com/cloudflare/wrangler2/pull/1027


SferaDev (approved):
LGTM! Who would've said that in 2022 we still use regex expressions to look for function usages in an AST

PR #263, opened 2022-06-06
CLI: random-data

gimenete:


SferaDev (approved):
Being playing with it, and it just works!!

Something minor I found (but it's also present in other commands) is that it requires me to have the env variable set. Although the "enquiry" has already asked me the workspace... Happens also for the ``shell`` command with the apiKey.

Screenshot 2022-06-07 at 12 22 10

Also I think we will eventually move this to a package for usage in the browser

gimenete:
@SferaDev great catch! I've fixed it in https://github.com/xataio/client-ts/pull/263/commits/875dee650d0dbdbb9824c60653a60b74e5710601

However for the shell I haven't been able to reproduce it.

Issue #191, opened 2022-05-14
Variadic types for return vs create


SferaDev:
When we add Dates support we can consider:

ColumnType<Date, Date | string>

Where we type on fn returns Date. But also accept strings in create/update.

Similarly we can have a third generic for the update:

ColumnType<string, string, never>

Which is useful for IDs or protected fields. Ie. when we add multi field primary identifiers.

The spec would be:

ColumnType<Return, Create = Return, Update = Create>

richardgill:
@SferaDev has this already been done. If not I've marked as P2 - please feel free to change it if that's not right.

PR #265, opened 2022-06-07
Support Date objects

SferaDev:
- Make client to parse and return "Date" objects for ``datetime`` fields.
- Do not depend on codegen ``links`` for schema retrieval
- Update unit tests to use ``msw``

Future improvement:

- Accept ``Date | string`` on create/update operations https://github.com/xataio/client-ts/issues/191


test/integration.test.ts
```diff
@@ -775,21 +775,21 @@ describe('getBranch', () => {
 
     const getBranchOptions = { apiKey: '', apiUrl: '', fetchImpl: {} as FetchImpl };
 
-    process.env = { XATA_BRANCH: branchName };
+    process.env = { NODE_ENV: 'development', XATA_BRANCH: branchName };
```
gimenete, line 778: Just curious. Why do we need this env now?

test/integration/dates.test.ts
```diff
@@ -73,7 +73,7 @@ describe('dates', () => {
     const date = new Date();
     const record = await xata.db.datetime.create({ date });
 
-    // TODO: Parse date in the client
-    expect(record.date).toEqual(date.toISOString());
+    expect(typeof record.date).toEqual('object');
```
gimenete, line 76: What do you think?

```suggestion
    expect(record.date instanceof Date).toEqual(true);
```

gimenete (approved):
LGTM! One question and one suggestion.

test/integration.test.ts
SferaDev: We don't for the tests to pass, but ``tsc`` complains that ``NodeEnv`` always needs to be set. Adding it to fix type errors from the compiler

PR #267, opened 2022-06-07
CLI: Generate API keys through a web flow

gimenete:
Depends on https://github.com/xataio/frontend-next/pull/887

It is more secure and easier for the user.


SferaDev (approved):
Works like a charm

PR #270, opened 2022-06-08
Move chunk to options object

SferaDev:
Closes: #269

It is very confusing that all methods have options as first and unique parameter, but getIterator/getAll have ``chunk, options``


gimenete:
👍 makes sense to put it in an options object.

What do you think about renaming it to `batchSize`?

gimenete (approved):

PR #286, opened 2022-06-10
CLI: Schema dump

gimenete:
```bash
# Print schema to stdout
xata schema dump
# Write schema to a file
xata schema dump -f schema.json
```

### DIFF
--- a/cli/src/commands/schema/dump.ts
+++ b/cli/src/commands/schema/dump.ts
@@ -11,7 +11,6 @@ export default class SchemaDump extends BaseCommand {
   static examples = [];
 
   static flags = {
-    ...this.commonFlags,
     file: Flags.string({ char: 'f', description: 'File to write the schma to' })
   };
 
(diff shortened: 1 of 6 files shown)


cli/src/commands/schema/dump.ts
```diff
@@ -0,0 +1,35 @@
...
+    file: Flags.string({ char: 'f', description: 'File to write the schma to' })
+  };
+
+  static args = [];
+
+  async run(): Promise<Schemas.Schema | undefined> {
+    const { flags } = await this.parse(SchemaDump);
+
+    const { databaseURL } = await this.getParsedDatabaseURL();
+    const branchDetails = await getCurrentBranchDetails({ fetchImpl: fetch, databaseURL });
```
SferaDev, line 23: Another one for https://github.com/xataio/client-ts/pull/291 😄

SferaDev (approved):

cli/src/commands/schema/dump.ts
SferaDev: We should change this to:

```suggestion
    const branchDetails = await this.getBranchDetails(databaseURL);
```

Depends which one hets merged before to apply the change.

gimenete: Hahaha. Nice catch. I think I'm going to merge this one first. We can fix this in your PR or another one.

PR #287, opened 2022-06-10
CLI: schema merge command

gimenete:
https://user-images.githubusercontent.com/50486/173024913-fc603fe9-49c7-4773-9bff-e4c1012d5b72.mp4


cli/src/commands/merge/index.ts
```diff
@@ -0,0 +1,36 @@
...
+import Codegen from '../codegen/index.js';
+
+export default class Merge extends BaseCommand {
+  static description = 'Merge the current branch with another branch';
+
+  static examples = [];
+
+  static flags = {};
+
+  static args = [{ name: 'branch', description: 'The branch to merge the current branch with', required: true }];
```
SferaDev, line 14: I'd probably add a way to ``xata merge foo-a foo-b``, not sure about the order (the second one is optional of course).

Also maybe an interactive way could be great: ``xata merge`` and it asks both branches origin/destination. Maybe destination with feault to ``main`` branch.

This can go to the improvements card.

cli/src/commands/merge/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  async run(): Promise<Schemas.Schema | undefined> {
+    const { args } = await this.parse(Merge);
+    const { branch } = args;
+
+    const { workspace, database, databaseURL } = await this.getParsedDatabaseURL();
+    const xata = await this.getXataClient();
+    const currentBranchDetails = await getCurrentBranchDetails({ fetchImpl: fetch, databaseURL });
+    if (!currentBranchDetails) return this.error('Could not resolve the current branch');
```
SferaDev, line 23: Needs to be updated after 😄

SferaDev (approved):
I'm good with it as-is. We can add the merge A B in the future.

PR #291, opened 2022-06-10
Fix some commands to read file api key

SferaDev:


SferaDev:
It works fine locally, but attempted to use npx and fails there "error performing authentication"

gimenete (commented):
An alternate implementation would be to have a `getCurrentBranchDetails` method in the base class for our commands that wraps the `getCurrentBranchDetails` from xata-client.

SferaDev:
@gimenete Updated by moving ``getBranchDetails`` to base command. Also added the ``get/create`` branch utility methods, as branch resolution PR is blocked. The existing ``getCurrentBranchDetails`` in ``config.ts`` can be probably remvoed as soon as we remove the ``generateFromContext`` in the old ``codegen`` package-level CLI command.

gimenete (approved):
✨  Nice!!

PR #292, opened 2022-06-11
Move shell to CLI

gimenete:
What the title says.

This PR also fixes the authentication problem for the shell if you don't have a XATA_API_KEY as env variable.

I think we should publish a last version of the shell package that prints a warning when executed telling the user to use the new CLI. Thougths?


SferaDev:
I would personally make the old package return a message suggesting to use the new CLI

cli/src/commands/shell/index.ts
```diff
@@ -1,11 +1,160 @@
...
+  };
+
   async run(): Promise<void> {
-    const databaseURL = await this.getDatabaseURL();
-    await run({ databaseURL });
+    const { flags } = await this.parse(Shell);
+    const apiKey = await readAPIKey();
+    if (!apiKey)
+      this.error('No API key found. Either use the XATA_API_KEY environment variable or run `xata auth login`');
+    const { protocol, host, databaseURL } = await this.getParsedDatabaseURL();
```
SferaDev, line 35: We're not using the ``branch``, we should review this as I think it will error out or use default branch even if we're in another one.

cli/src/commands/shell/index.ts
```diff
@@ -1,11 +1,160 @@
...
+
+    this.log(
+      `Connected to ${chalk.bold(databaseURL)}. There's a XataClient instance in the ${chalk.bold(
+        'xata'
+      )} global variable.`
+    );
+
+    const fetchApi = async (method: string, path: string, body?: any) => {
+      // TODO: Add support for staging, how?
+      const baseUrl = path.startsWith('/db') ? `${protocol}//${host}` : 'https://api.xata.io';
```
SferaDev, line 56: We should probably use the helper methods of the base command here

cli/src/commands/shell/index.ts
```diff
@@ -1,11 +1,160 @@
...
+    const apiKey = await readAPIKey();
+    if (!apiKey)
+      this.error('No API key found. Either use the XATA_API_KEY environment variable or run `xata auth login`');
+    const { protocol, host, databaseURL } = await this.getParsedDatabaseURL();
+
+    // Generate the file in the same dir than this package's code so it
+    // can import @xata.io/client
+    const tempFile = path.join(__dirname, `xata-${Date.now()}.mjs`);
+    try {
+      const { transpiled } = await generateFromContext('javascript', { apiKey, databaseURL });
```
SferaDev, line 41: I believe ``generateFromContext`` should be deprecated in the ``codegen`` package, maybe not in this PR but in the near future.

cli/src/commands/shell/index.ts
gimenete: What do you think about https://github.com/xataio/client-ts/pull/292/commits/d8f59a1db34e6b54122c59809b4a19121cf147d4

gimenete: What would you use in this case?

cli/src/commands/shell/index.ts
gimenete: Yup. Let's do it in another PR.

packages/shell/src/cli-run.ts
```diff
@@ -1,162 +1,5 @@
...
-import { homedir, tmpdir } from 'os';
-import path, { dirname } from 'path';
-import RJSON from 'relaxed-json';
-import repl from 'repl';
-import { fileURLToPath } from 'url';
-
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = dirname(__filename);
-
 export async function run(options: { env?: string; databaseURL?: string; apiKey?: string; code?: string } = {}) {
```
SferaDev, line 1: Do we still need this file?

cli/src/commands/shell/index.ts
SferaDev: Cool!

cli/src/commands/shell/index.ts
SferaDev: Not sure, but I would like to have ``xata login auth --staging`` or ``xata login auth --backend http://localhost:5001`` for example, and that it remembers it (in config). So in this particular case it knows through a base command helper which API to use.

packages/shell/src/cli-run.ts
gimenete: The arguments? Whatever you prefer. I think keeping it and throwing an error is fine.

cli/src/commands/shell/index.ts
gimenete: The old xata CLI uses an env variable. Maybe we could do the same. What I don't like is storing an API key without specifying the environment somehow. Maybe we can convert the `key` file into an `.ini` file? And having a section per environment?

cli/src/base.ts
```diff
@@ -278,21 +283,22 @@ export abstract class BaseCommand extends Command {
...
+        branch = await getCurrentBranchName({
+          fetchImpl: fetch,
+          databaseURL: info.databaseURL,
+          apiKey: (await readAPIKey()) ?? undefined
+        });
+      } else if (process.env.XATA_BRANCH !== undefined) {
+        branch = process.env.XATA_BRANCH;
+      } else {
+        branch = await this.getBranch(info.workspace, info.database);
+      }
```
SferaDev, line 301: Maybe simplify the if condition?

```suggestion
    let branch = '';

    if (branchFlag) {
      branch = branchFlag;
    } if (info.source === 'config') {
      branch = await getCurrentBranchName({
        fetchImpl: fetch,
        databaseURL: info.databaseURL,
        apiKey: (await readAPIKey()) ?? undefined
      });
    } else if (process.env.XATA_BRANCH !== undefined) {
      branch = process.env.XATA_BRANCH;
    } else {
      branch = await this.getBranch(info.workspace, info.database);
```

packages/shell/src/cli-run.ts
SferaDev: Sorry, no. I meant if the cli-run and cli both are needed. Both are right now throwing errors, but I guess only one is being executed (cli one) and the other one is left unused (cli-run).

cli/src/commands/shell/index.ts
SferaDev: That would be great to be honest.

SferaDev (approved):

cli/src/base.ts
```diff
@@ -278,12 +283,14 @@ export abstract class BaseCommand extends Command {
...
-  async getParsedDatabaseURLWithBranch(databaseURLFlag?: string, allowCreate?: boolean) {
+  async getParsedDatabaseURLWithBranch(databaseURLFlag?: string, branchFlag?: string, allowCreate?: boolean) {
     const info = await this.getParsedDatabaseURL(databaseURLFlag, allowCreate);
 
     let branch = '';
 
-    if (info.source === 'config') {
+    if (branchFlag) {
+      branch = branchFlag;
+    } if (info.source === 'config') {
```
SferaDev, line 293:
```suggestion
    } else if (info.source === 'config') {
```

packages/shell/src/cli-run.ts
gimenete: The reason I kept `cli-run` is because it is imported by `index.ts`, which is the file that is exported as a library. So I'm keeping both and one will fail when used as a CLI app and the other one will fail when used as a library.

PR #297, opened 2022-06-13
CLI init fixes

gimenete:
This PR fixes

- Error when using the codegenerator and a JS file.
- Ctr+C when asked if installing the SDK or using the code generator.

### DIFF
--- a/cli/src/commands/init/index.ts
+++ b/cli/src/commands/init/index.ts
@@ -89,6 +89,7 @@ export default class Init extends BaseCommand {
       message: 'Do you want to use the TypeScript/JavaScript code generator?',
       initial: true
     });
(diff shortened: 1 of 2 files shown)


cli/src/commands/init/index.ts
```diff
@@ -75,6 +75,7 @@ export default class Init extends BaseCommand {
       message: 'Do you want to install the TypeScript/JavaScript SDK?',
       initial: true
     });
+    if (confirm === undefined) return this.exit(1);
```
SferaDev, line 78: We should add a linter for this

gimenete: Yeah, or a method in the base class I guess.

SferaDev (approved):

PR #300, opened 2022-06-14
Update linter rules

SferaDev:
Add some more stricter rules. Only in the client SDK for now.


packages/client/src/client.ts
```diff
@@ -39,7 +40,7 @@ export const buildClient = <Plugins extends Record<string, XataPlugin> = {}>(plu
       this.search = search;
 
       for (const [key, namespace] of Object.entries(plugins ?? {})) {
-        if (!namespace) continue;
+        if (namespace == undefined) continue;
```
SferaDev, line 43:
```suggestion
        if (namespace === undefined) continue;
```

packages/client/.eslintrc.cjs
```diff
@@ -6,8 +6,7 @@ module.exports = {
     project: 'packages/client/tsconfig.json'
   },
   rules: {
-    '@typescript-eslint/no-explicit-any': 'off',
-    '@typescript-eslint/ban-types': 'off',
     '@typescript-eslint/no-floating-promises': 'error',
+    "@typescript-eslint/strict-boolean-expressions": ["error", { allowNullableString: true, allowNullableObject: true }],
```
SferaDev, line 10: Allow Nullable String is not safe for empty strings. Might revisit in the future.

gimenete (approved):
Nice!!

PR #301, opened 2022-06-14
CLI profiles

gimenete:
Instead of storing just a single API key, this PR allows the CLI to store "profiles". Users for now will only use one profile (named `profile`), but this allows:

- For us, storing also a `host` parameter and have different profiles for different environments.
- In the future exposing the profiles feature to users and allow them to use different accounts, or different auth mechanisms.

Since for now we don't need to expose this to users, the way to specify the profile to use is using an env variable. In the future we can add flags all/most commands.

This is a small breaking change because the old API key file is not used anymore.

An example file would be:

```
[default]
apiKey=xau_1234abcdef
```

This PR is inspired on what AWS does https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html


cli/src/commands/auth/login.ts
```diff
@@ -38,7 +38,7 @@ export default class Login extends BaseCommand {
 
     await this.verifyAPIKey(key);
 
-    await writeAPIKey(key);
+    await setProfile({ apiKey: key });
```
SferaDev, line 41: Auth login command hangs:

image

cli/src/commands/auth/login.ts
```diff
@@ -1,7 +1,7 @@
 import prompts from 'prompts';
 import { createAPIKeyThroughWebUI } from '../../auth-server.js';
 import { BaseCommand } from '../../base.js';
-import { readAPIKeyFromFile, writeAPIKey } from '../../key.js';
+import { getProfile, setProfile } from '../../credentials.js';
 
 export default class Login extends BaseCommand {
   static description = 'Authenticate with Xata';
```
SferaDev, line 7: Unrelated to this PR (actually error in the frontend). If key name already exists, strange error:

image

cli/src/commands/auth/login.ts
gimenete: Yeah, it takes a couple of seconds to close existing http connections because the http server is gracefulled closed.

I've pushed this change https://github.com/xataio/client-ts/pull/301/commits/002b8408d2c9f25e52964b7c78efc83d7eafb632 and should be immediate now.

cli/src/commands/auth/login.ts
gimenete: Good to know. I'll investigate 👍

gimenete: PR to fix that https://github.com/xataio/frontend-next/pull/914

SferaDev (approved):
LGTM. Let's iterate with multiple hosts in another PR

PR #309, opened 2022-06-16
Improvements to raw api shell command

SferaDev:
- Updates to raw api shell
- Fix pretty print hiding objects at some depth
- Check for object type

### DIFF
--- a/cli/src/commands/shell/index.ts
+++ b/cli/src/commands/shell/index.ts
@@ -57,6 +57,14 @@ export default class Shell extends BaseCommand {
       // TODO: Add support for staging, how?
       const baseUrl = path.startsWith('/db') ? `${protocol}//${host}` : 'https://api.xata.io';
 
+      const parse = () => {
+        try {
+          return JSON.stringify(RJSON.parse(body));
+        } catch (error) {
+          return undefined;
+        }
+      };
+
       try {
         const response = await fetch(`${baseUrl}${path}`, {
           method,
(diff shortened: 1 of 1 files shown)


cli/src/commands/shell/index.ts
```diff
@@ -111,6 +120,11 @@ export default class Shell extends BaseCommand {
 
     replServer.context.xata = new XataClient({ fetch, apiKey });
     replServer.context.api = new XataApiClient({ fetch, apiKey });
+    replServer.context.api.GET = (path: string) => fetchApi('GET', path);
+    replServer.context.api.POST = (path: string, body?: any) => fetchApi('POST', path, JSON.stringify(body));
+    replServer.context.api.PATCH = (path: string, body?: any) => fetchApi('PATCH', path, JSON.stringify(body));
+    replServer.context.api.PUT = (path: string, body?: any) => fetchApi('PUT', path, JSON.stringify(body));
+    replServer.context.api.DELETE = (path: string) => fetchApi('DELETE', path);
```
tsg, line 127: Nice! Thanks :)

Would it be possible to do the same in the SDK playground?

tsg (approved):
Awesome, thanks! I just left a question.

gimenete (approved):
😍  nice!!

cli/src/commands/shell/index.ts
SferaDev: Then I'd move it to another place

PR #312, opened 2022-06-16
Use new table search with filters

SferaDev:
``api.records.searchTable(workspace, database, branch, table, { query: "foo", filter: { name: "bar" } })``

``xata.db.table.search("foo", { filter: { name: "bar" } })``

### DIFF
--- a/cli/src/commands/shell/index.ts
+++ b/cli/src/commands/shell/index.ts
@@ -96,7 +96,7 @@ export default class Shell extends BaseCommand {
           if (['get', 'post', 'patch', 'put', 'delete'].includes(verb.toLowerCase())) {
             const body = rest.join(' ');
             return fetchApi(verb, path, body).then((result) => {
-              return postProcess(result, { table }, callback);
+              return postProcess(result, { table, deep: true }, callback);
             });
           }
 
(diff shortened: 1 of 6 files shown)


packages/client/src/schema/repository.ts
```diff
@@ -453,12 +458,19 @@ export class RestRepository<Data extends BaseData, Record extends XataRecord = D
...
 
-    const { records } = await searchBranch({
-      pathParams: { workspace: '{workspaceId}', dbBranchName: '{dbBranch}' },
-      body: { tables: [this.#table], query, fuzziness: options.fuzziness },
+    const { records } = await searchTable({
+      pathParams: { workspace: '{workspaceId}', dbBranchName: '{dbBranch}', tableName: this.#table },
+      body: {
+        query,
+        fuzziness: options.fuzziness,
+        filter: options.filter as Schemas.FilterExpression
```
SferaDev, line 472: I need to battle with these types. Opening https://github.com/xataio/client-ts/issues/313

cli/src/commands/shell/index.ts
```diff
@@ -154,13 +154,17 @@ function getDefaultEval() {
...
+  options: { table: boolean; deep?: boolean },
+  callback: (err: Error | null, result: any) => void
+) {
   const { table } = options;
   return callback(
     null,
     table
       ? console.table(result)
-      : typeof result === 'object'
+      : typeof result === 'object' && options.deep
```
gimenete, line 167: Why are you adding the `deep` flag? Don't we want that not only in raw API calls, but in JS statements as well?

gimenete (approved):
LGTM. Just one question.

cli/src/commands/shell/index.ts
SferaDev: Tudor noticed that autocomplete is also using postprocessing and had some weird behavior. I'll forward you the video.

Added it as an opt-in flag, and let's decide where to add it.

Issue #314, opened 2022-06-20
Page offset must not exceed 800


SferaDev:
<img width="567" alt="Screenshot 2022-06-20 at 11 03 12" src="https://user-images.githubusercontent.com/2181866/174566240-e5ada117-e8b0-4971-a0dd-24d2db84bd17.png">

Repro: Create bulk 10k items, fails when retrieving the objects

PR #325, opened 2022-06-22
Use cursor pagination with many operations

SferaDev:
Closes: #314 

- Change internal methods to use cursor navigation instead of offset navigation
- Behaviour change: Ignore ``sorting/filtering`` when sending cursor navigation.
  - Backend would fail with that request
  - In userland TypeScript type guards do not allow it
  - Our internal methods could still generate bad queries
  - Ignore them for our internal methods and JavaScript users

Regarding this behaviour change, I have mixed feelings to be honest. I kind of like that we do not error out on impossible situations, but could mislead a JavaScript user who doesn't have type support.

### DIFF
--- a/test/integration.test.ts
+++ b/test/integration.test.ts
@@ -2,16 +2,15 @@ import { execSync } from 'child_process';
 import fetch from 'cross-fetch';
 import dotenv from 'dotenv';
 import { join } from 'path';
-import { afterAll, beforeAll, describe, expect, test, vi } from 'vitest';
+import { afterAll, beforeAll, describe, expect, test } from 'vitest';
 import { BaseClient, contains, isXataRecord, lt, Repository, XataApiClient } from '../packages/client/src';
-import { FetchImpl } from '../packages/client/src/api/fetcher';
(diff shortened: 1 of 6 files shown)


test/integration.test.ts
```diff
@@ -266,7 +265,7 @@ describe('integration tests', () => {
     expect(page2And3.records).toEqual([...page2.records, ...page3.records]);
   });
 
-  test('fails if sending cursor with sorting', async () => {
+  test('we ignore sorting if sending it with cursor', async () => {
```
SferaDev, line 268: This is the behaviour change we are introducing for JS users

gimenete (commented):
> Regarding this behaviour change, I have mixed feelings to be honest. I kind of like that we do not error out on impossible situations, but could mislead a JavaScript user who doesn't have type support.

Yeah. I'd prefer to throw than having unexpected behavior that is hard to debug.

The cursor pagination changes LGTM.

SferaDev:
> Yeah. I'd prefer to throw than having unexpected behavior that is hard to debug.

Fixed!

gimenete (approved):
🚀

Issue #326, opened 2022-06-22
Watch mode for codegen


TejasQ:
Just adding this so we don't lose track of it

please ignore for now

SferaDev:
Watch mode for codegen, right?

TejasQ:
Yes sir

SferaDev:
PR for testing, every second request. We will need websocket support for a better approach, for now if you need it for demos you have it around. I'll let @gimenete decide if we merge as-is or not.

PR #328, opened 2022-06-22
Add watch mode for the codegen

SferaDev:
Closes: #326 

Also bring back ``out`` flag for codegen

### DIFF
--- a/.changeset/dull-rules-do.md
+++ b/.changeset/dull-rules-do.md
@@ -1,0 +1,5 @@
+---
+'@xata.io/cli': patch
+---
+
+Add watch mode for codegen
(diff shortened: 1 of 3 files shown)


gimenete (commented):
I think the watch mode made sense when we had the local schema file. But that's not true anymore. I see little value having a watch mode and could potentially generate a lot of traffic to our servers.

SferaDev:
I agree it's a greedy operation right now, with proper backend support we could have something better.

In any case even if we don't have schema files locally anymore, I see it useful if you're interacting with the schema on the webpage and you receive changes in the moment to your code.

Some kind of watch mode will be needed in the CLI, and the VSCode extension will definitely have it too.

We can raise it up maybe in the next frontenders sync.

PR #329, opened 2022-06-22
Update codegen instructions

SferaDev:


README.md
```diff
@@ -55,20 +55,18 @@ await client.workspaces.deleteWorkspace(workspace);
...
-  }
-}
+To run it (and assuming you have configured the project with `xata init`):
+
+```bash
+xata codegen
 ```
 
-Run it (this is assuming your Xata directory is in `./xata` and it was created by the `xata` cli application)
+If you want to override the output file, you can pass the ``--out`` flag with the path to the output file.
```
SferaDev, line 66: This flag was removed, but I'm planning to re-add it with #328

README.md
```diff
@@ -10,7 +10,7 @@ Install the dependencies
 
 ```
 npm install @xata.io/client
-npm install @xata.io/codegen -D
+npm install @xata.io/cli -g
```
richardgill, line 13: I was expecting to see `-D` still for this? Since if a teammate checks out your project it wouldn't work with `-g`.

README.md
richardgill: When I just used `xata init` it added 

```  
  "codegen": {
    "output": "./db/xata.ts"
  }
```

to my `.xatarc`.

Would/should I pass the `--out` argument?

README.md
```diff
@@ -55,20 +55,18 @@ await client.workspaces.deleteWorkspace(workspace);
...
 
-```json
-{
-  "scripts": {
-    "xata-codegen": "xata-codegen xata -o src/xata.ts"
-  }
-}
+To run it (and assuming you have configured the project with `xata init`):
+
+```bash
```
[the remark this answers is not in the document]
SferaDev: You could still have:

```json
"scripts": {
  "xata-codegen": "npx @xata.io/cli codegen"
}
```

But it's not the default way or recommended. We should still show it as an option in the docs.

README.md
SferaDev: Not needed, it is optional if you want to override it. Maybe we can remove it from the docs, they can figure it out with `--help`

gimenete (approved):

README.md
SferaDev, line 66:
```suggestion
```

PR #340, opened 2022-06-23
Delete shell package

gimenete:
It is no longer needed since we have it in the CLI and it's not available as a standalone package.

Right now the latest version throws if you try to use it, telling you to use the CLI instead.


SferaDev (approved):
Should we run ``npm deprecate``?

gimenete:
@SferaDev great idea. I'll do that 👍

PR #341, opened 2022-06-23
Deprecate codegen cli

gimenete:
We have now that functionality in the CLI.

This PR also changes a couple of CLI commands to not use `generateFromContext` anymore because this method is deleted too.


cli/src/commands/codegen/index.ts
```diff
@@ -17,17 +17,29 @@ export default class Codegen extends BaseCommand {
   static examples = [];
 
   static flags = {
-    ...this.commonFlags
+    ...this.commonFlags,
+    databaseURL: this.databaseURLFlag,
+    branch: this.branchFlag,
+    output: Flags.string({
+      name: 'output',
```
SferaDev, line 24: Oclif.Flags does not seem to have a property called `name`. Changing to `o` shorthand.

```suggestion
      char: 'o',
```

cli/src/commands/codegen/index.ts
```diff
@@ -17,17 +17,29 @@ export default class Codegen extends BaseCommand {
...
+      description: 'Output file. Overwrites your project configuration setting'
+    }),
+    declarations: Flags.boolean({
+      name: 'declarations',
```
SferaDev, line 28:
```suggestion
```

SferaDev (approved):

Issue #365, opened 2022-06-27
`xata schema edit` exits when selecting "Table"


TejasQ:
## Status Quo

![2022-06-27 15 18 39](https://user-images.githubusercontent.com/9947422/175951414-cf84fef2-b98c-4f27-9852-7df0c5d4338f.gif)

There's a selectable "Tables" heading in the CLI. Selecting this exits the CLI.

## Expected Behavior

Tables is not selectable, but either the first table is focused, or "Add a table" is focused by default.

Issue #351, opened 2022-06-26
No shorthand flags


TejasQ:
The Xata CLI works great with long-form flags (like `--force`), but does not have equivalent short-form flags (`-f`). Let's fix this to improve the developer experience.

gimenete:
We do have some shorthands. Example:

https://github.com/xataio/client-ts/blob/4f36065a2dba5473143c61aa5792863fc4a3caa4/cli/src/commands/schema/dump.ts#L14

and 

https://github.com/xataio/client-ts/blob/4f36065a2dba5473143c61aa5792863fc4a3caa4/cli/src/commands/codegen/index.ts#L23-L26

I'll add some more, but most of the times I just prefer not having a shorthand if it's not clear enough.

TejasQ:
Yeah, from what I've spoken to developers they have mixed feelings about shorthands, but the often used ones are convenient (like `-d` for databaseUrl) or similar. Happy to trust your judgment here based on what _should_ and _shouldn't_ have shorthand versions and iterate as needed.

I apologies for not doing the legwork of listing _specific_ shorthand properties that would have helped our developer experience with the CLI. I'll be a bit more thorough next time.

Issue #348, opened 2022-06-26
CLI auth success page needs design


TejasQ:
After creating an API key, we show this to a user:

<img width="1892" alt="image" src="https://user-images.githubusercontent.com/9947422/175812964-fec4a281-10b3-4f02-a316-8f74dc4adb1d.png">

Let's make this look better.

cc @DavidCnoops for ideas

AlenaNik:
@TejasQ to keep the Empty / Error / Success pages consistent across app UI and SDK we are keeping it aligned to the top, and centered horizontally at the moment. The reason: on taller screen it might look a bit weird and 'disconnected' from the rest of the UI. On smaller screens (e.g. laptop size) we could align it to the center. However, we didn't make the decision yet and for SDK we can iterate a bit more 👍  cc @DavidCnoops

PR #367, opened 2022-06-27
More small CLI improvements

gimenete:
- Closes https://github.com/xataio/client-ts/issues/365
- Improves error message when there's no API key
- Closes https://github.com/xataio/client-ts/issues/351
- Improves https://github.com/xataio/client-ts/issues/348 See screenshot

image


SferaDev (dismissed):

TejasQ:
> Closes https://github.com/xataio/client-ts/issues/348 See screenshot

This does not close the issue because there has not been UX/UI input. Sure, we've tweaked the UI a bit and it's a great improvement, but it still does not look finished.

TejasQ (changes requested):
Seems the canary build fails?

cli/src/api-key-success.html
```diff
@@ -0,0 +1,39 @@
...
+          'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+
+        text-align: center;
+        margin-top: 48px;
+      }
+    </style>
+  </head>
+  <body>
+    <div style="width: 200px; margin: 0 auto">
+      <svg viewBox="0 0 143 41" focusable="false" class="chakra-icon chakra-icon css-1aptyfp">
```
AlenaNik, line 16:
```
<svg aria-hidden="true" focusable="false">
  <!-- svg content -->
</svg>
```

would make more sense, focusable="false" is just IE hack so we also need to hide the icon with aria-hidden as it's pure decorative.

The other option:
```
<svg class="critter" viewBox="0 0 100 100" aria-labelledby="uniqueTitleID" aria-describedby="uniqueDescID" role="img">
  <title id="uniqueTitleID">Baxter</title>
  <desc id="uniqueDescID">A happy little bear</desc>
</svg>
```

TejasQ:
I wonder if it's time for automated test runs on each command.

cli/src/base.ts
```diff
@@ -88,10 +88,14 @@ export abstract class BaseCommand extends Command {
     const profile = apiKey ? undefined : await getProfile();
 
     apiKey = apiKey || profile?.apiKey;
-    if (!apiKey)
-      this.error(
-        'Could not instantiate Xata client. No API key found. Please run `xata auth login` or configure a project with `xata init`.'
-      );
+    if (!apiKey) {
+      this.error('Could not instantiate Xata client. No API key found.', {
+        suggestions: [
```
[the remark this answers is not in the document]
gimenete: Maybe. I discovered it the other day :)

I'll do some polishing of most commands (like adding examples and such) and I'll probably check if we can use it in some places. It also has a `ref` thing to put a URL, that's also a bit nice.

gimenete:
I removed the `Closes` word. I think this should unblock this PR in that regard. There's no urgency for a more polished design, especially when the current implementation is a plain text message.

In another PR we can implement a design from @DavidCnoops or @AlenaNik

cli/src/api-key-success.html
gimenete: Cool. I've added `aria-hidden` here https://github.com/xataio/client-ts/pull/367/commits/72a8371446c35baf4e19367f58831eba7a2dbd9a based on your first suggestion.

cli/src/base.ts
```diff
@@ -43,12 +43,12 @@ export abstract class BaseCommand extends Command {
   searchPlaces = [`.${moduleName}rc`, `.${moduleName}rc.json`, 'package.json'];
 
   static databaseURLFlag = Flags.string({
-    name: 'databaseurl',
+    char: 'u',
```
TejasQ, line 46: Why not `d`?

gimenete: Because it's a **U**rl and `-d` is used frequently as a _delete_ flag. E.g. `git branch -d to-be-deleted`.

TejasQ: I get that `u` means `url` (maybe?), but since it starts with `d`, I expect the shorthand to also be `d`. Even the help page looks a bit more confusing with `-u`:

image

cli/src/base.ts
```diff
@@ -43,12 +43,12 @@ export abstract class BaseCommand extends Command {
...
     description: 'URL of the database in the format https://{workspace}.xata.sh/db/{database}'
   });
 
   static branchFlag = Flags.string({
-    name: 'branch',
+    char: 'b',
```
TejasQ, line 51: Sorry if this is a silly question, but is this shorthand used? The help page doesn't show it:

image

gimenete:
@TejasQ @SferaDev I've fixed the canary build. There was a missing dependency. I've added a changeset to bump the version.

cli/src/base.ts
gimenete: Fixed here https://github.com/xataio/client-ts/pull/367/commits/116ea09970c6e51ddebf122deb64a152305cafe4

cli/src/base.ts
```diff
@@ -88,10 +88,14 @@ export abstract class BaseCommand extends Command {
     const profile = apiKey ? undefined : await getProfile();
 
     apiKey = apiKey || profile?.apiKey;
-    if (!apiKey)
-      this.error(
-        'Could not instantiate Xata client. No API key found. Please run `xata auth login` or configure a project with `xata init`.'
-      );
+    if (!apiKey) {
+      this.error('Could not instantiate Xata client. No API key found.', {
```
TejasQ, line 92: Should this error? Or should it call the login command directly?

cli/src/base.ts
TejasQ: Right, `-d` to delete makes sense. I would still prefer to go with `-d` for now and iterate, possibly changing it to `-u` based on feedback.

gimenete: > git, -D deletes instead of -d

With git `-d` deletes as well. LIke in the example I left above.

TejasQ: Right, I'm aware. I would still prefer to go with `-d` for now and iterate, possibly changing it to `-u` based on real-world developer feedback. Can we do this?

gimenete: I prefer not having a shorthand than having a shorthand that can potentially change. I don't want to break users workflows.

cli/src/base.ts
gimenete: I think erroring is fine.

cli/src/base.ts
tsg: Would a two char short version, like `-db` be possible? I've noticed they do this in npm, for example: `[-ws|--workspaces]`

gimenete: I'm afraid that's not possible. Oclif asks for a single char.

image

We could change `databaseURL` to `db`, but not sure if that's clear enough.

cli/src/base.ts
SferaDev: Agreed, I wouldn't call another command at least without asking. The `suggestions` helper looks really nice in `this.error`.

cli/src/base.ts
SferaDev: There's an `enhacement` issue for aliases: https://github.com/oclif/oclif/issues/438

What about duplicating the flag and reading both (databaseURL and db)?

I kind of agree it's a long one

SferaDev (approved):
LGTM. We can address the UI redesign in another PR.

cli/src/base.ts
TejasQ: Sure, `--db` is much clearer than `-u` and also shorter than `--databaseURL`, which will probably be confused with `--databaseUrl` quite often as well.

gimenete: I'm going to remove the shorthand and will open a different PR for this specific flag.

gimenete: New PR https://github.com/xataio/client-ts/pull/368

PR #368, opened 2022-06-28
Change --databaseURL flag to --db

gimenete:
See https://github.com/xataio/client-ts/pull/367#discussion_r908325418

I've decided to change the way we use it internally to:

```diff
  static flags = {
      ...this.commonFlags,
+     ...this.databaseURLFlag
-     databaseURL: this.databaseURLFlag
    };
```

I'll do the same with the other common flags. The benefit is that the name of the flag is not repeated.

Also in this PR I've put common flags in their own flag group.

### DIFF
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -44,6 +44,7 @@ export abstract class BaseCommand extends Command {
 
   static databaseURLFlag = {
     db: Flags.string({
+      helpValue: 'https://{workspace}.xata.sh/db/{database}',
       description: 'URL of the database in the format https://{workspace}.xata.sh/db/{database}'
     })
   };
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -42,10 +42,11 @@ export abstract class BaseCommand extends Command {
   // In the future we can support YAML
   searchPlaces = [`.${moduleName}rc`, `.${moduleName}rc.json`, 'package.json'];
 
-  static databaseURLFlag = Flags.string({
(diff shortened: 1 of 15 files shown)


cli/src/base.ts
```diff
@@ -27,6 +27,8 @@ const partialProjectConfig = projectConfigSchema.deepPartial();
 export type ProjectConfig = z.infer<typeof partialProjectConfig>;
 
 const moduleName = 'xata';
+const commonFlagsHelpGroup = 'Common';
```
SferaDev, line 30: Really cool

cli/src/base.ts
```diff
@@ -42,20 +44,26 @@ export abstract class BaseCommand extends Command {
   // In the future we can support YAML
   searchPlaces = [`.${moduleName}rc`, `.${moduleName}rc.json`, 'package.json'];
 
-  static databaseURLFlag = Flags.string({
-    description: 'URL of the database in the format https://{workspace}.xata.sh/db/{database}'
-  });
+  static databaseURLFlag = {
+    db: Flags.string({
+      helpValue: 'https://{workspace}.xata.sh/db/{database}',
+      description: 'URL of the database in the format https://{workspace}.xata.sh/db/{database}'
```
SferaDev, line 50: I find the format redundant right now. We show it both in the description and help value. 


```suggestion
      helpValue: 'https://{workspace}.xata.sh/db/{database}',
      description: 'URL of the database'
```

SferaDev (approved):

PR #369, opened 2022-06-28
Illustration and design after an API key has beeen created successfully

gimenete:
Closes https://github.com/xataio/client-ts/issues/348

image


TejasQ:
Question for @xataio/designers: https://github.com/xataio/client-ts/issues/348#issuecomment-1168695800

SferaDev:
I'll leave the design approval for @xataio/designers 

Something I found though while testing it on a headless SSH session is that we don't show the url at all.

In my opinion we should either always show the url so it can be copy pasted, or at least detect the open() has failed and show it then.

gimenete:
> at least detect the open() has failed and show it then

Yup. We are doing that already!

SferaDev:
> > at least detect the open() has failed and show it then
> 
> Yup. We are doing that already!

Weird didn't work for me. I couldn't finish the auth flow through ssh. Note: I do have browsers installed, just connected to a session without an XServer.

gimenete:
Makes sense. I'll do a PR for that

SferaDev (approved):

AlenaNik (approved):

PR #372, opened 2022-06-28
Docs docs docs

gimenete:
Better Readme files.


cli/README.md
```diff
@@ -1 +1,74 @@
...
+```bash
+npm install @xata.io/cli --save-dev
+```
+
+## Authentication, API keys and projects
+
+Almost all commands can work in two different modes: globally or scoped to a project. You can configure a project with `xata init`. This will create a `.xatarc` file (although `.xatarc.json` and `package.json` are also supported) and an `.env` file.
+
+- The project configuration file contains some preferences and information, including the `databaseURL`.
+- The `.env` file contains an API key and optionally a fallback branch name. More about this later.
```
SferaDev, line 24: I'd personally remove the "more about this later".

Also maybe add here a mention that this file shouldn't be checked out into git. Maybe you added it after, doing a linear reading. In any case, I think it would be worth mentioning that now.

Question: We would probably want to git check out the fallback branch. Why didn't we add this to the project config?

cli/README.md
```diff
@@ -1 +1,74 @@
...
+
+- It will create a project configuration file (`.xatarc`).
+- It will create or update your `.env` file to store a database-scoped API key. (At the time of writing these docs it'll actually store a personal API key, but this will change soon).
+- It will ask you if you want to install the SDK and/or use the TypeScript/JavaScript code generator.
+
+## Code generation
+
+You can use the Xata SDK (`@xata.io/client`) directly or with code generation. The benefit of using code generation is that you'll have a Xata client with specific types and classes for your database schema. This is great for TypeScript because it ensures type safety, but also for JavaScript because you can benefit from its type definitions if you use development tools such as IDEs or text editors that support them.
+
+Code generation will be executed automatically in some cases, such as when creating a new Xata branch, but you can also invoke it directly with `xata codegen` if you've done change to the schema through the UI or the REST API and you want to update your local generated code.
```
SferaDev, line 44:
```suggestion
Code generation will be executed automatically in some cases by the CLI (such as when creating a new Xata branch). You can always invoke it directly with `xata codegen`.

If you make changes to the schema through the UI or the REST API, you will need to run `xata codegen` to update your local generated code with the new changes.

The codegen command completely overwrites the contents of the output file. Any changes you make to the client file will be likely lost in a code generation change.
```

SferaDev (approved):
Changes look good to me, we will iterate a lot on these. I wouldn't consider them final yet, but it's a great starting point. Instead of blocking you here I think we can propose more content and changes in future PRs.

I think we should have sections in separate md files that the docs can read. Maybe even MDX or similar to read JSON files and add things from the code. We should probably have a meeting with DevRel soonish to find a good approach that we can easily maintain.


I want to also redo some parts of the SDK docs. Mainly:

- Split into sections leaving only the simple usage in the main one
- Add a section about Plugins
- Add a section about the Cache
- Rewrite the API Client part to be used as a plugin
- Add more about advanced queries, composition (any/all) and how to run them.

cli/README.md
gimenete: > We would probably want to git check out the fallback branch. Why didn't we add this to the project config?

Because the fallback branch can be different between people in the same project. That's even the ideal scenario, because you don't want to share the same dataset with another teammate. That could be a bit chaotic.

TejasQ:
> We should probably have a meeting with DevRel soonish to find a good approach that we can easily maintain.

Thanks!

For the future, please always request a review from xataio/devrel on public facing things like this.

PR #375, opened 2022-06-29
Changes to getMany and getPaginated

SferaDev:
- Update docs to recommend `getPaginated` and `getAll` where needed
- Return new custom `RecordArray<T>` with pagination in `getMany`
- Add warning to `getMany` if there're more pages not fetched yet
- Update default pagination size from 200 to 20


packages/client/src/schema/query.ts
```diff
@@ -313,12 +314,17 @@ export class Query<Record extends XataRecord, Result extends XataRecord = Record
...
   getMany<Options extends RequiredBy<QueryOptions<Record>, 'columns'>>(
     options: Options
-  ): Promise<SelectedPick<Record, typeof options['columns']>[]>;
+  ): Promise<RecordArray<SelectedPick<Record, typeof options['columns']>>>;
+
+  async getMany<Result extends XataRecord>(options: QueryOptions<Record> = {}): Promise<RecordArray<Result>> {
+    const page = await this.getPaginated(options);
+
+    if (page.hasNextPage() && options.pagination?.size === undefined) {
+      console.debug('Calling getMany does not return all results. Paginate to get all results or call getAll.');
```
SferaDev, line 323: debug is less noisy, but does not provide the easiest way to jump from console as trace. If docs are well explained and we recommend getPaginated, this could be enough.

SferaDev, line 323:
```suggestion
      console.trace('Calling getMany does not return all results. Paginate to get all results or call getAll.');
```

gimenete (approved):
LGTM! Thanks for implementing these changes!

PR #380, opened 2022-07-01
Fix spawn usage on windows

fabien0102:
## Bug description
The `xata init` command was failing for me:

image

## Resolution
Adding `{ shell: true}` seams to do the job.

## Reference
https://stackoverflow.com/questions/60386867/node-spawn-child-process-not-working-in-windows

### DIFF
--- a/.changeset/tasty-nails-trade.md
+++ b/.changeset/tasty-nails-trade.md
@@ -1,0 +1,5 @@
+---
+"@xata.io/client": patch
+---
+
+Fix spawn usage on windows
--- a/.changeset/tasty-nails-trade.md
+++ b/.changeset/tasty-nails-trade.md
@@ -1,5 +1,5 @@
 ---
-"@xata.io/client": patch
+"@xata.io/cli": patch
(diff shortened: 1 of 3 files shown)


SferaDev:
@gimenete WDYT? Tested on windows and works fine now

gimenete:
It's weird that we need to use the `shell` for this. 🤔  We shouldn't, right? We are not using any functionality of a shell.

Also, it is failing on CI:

> Error: /bin/sh: 1: Syntax error: "(" unexpected

gimenete:
Maybe we need the full path of the binary and/or the .exe extension on windows?

tsg:
Hmm, yeah, if we could avoid `shell: true` it would be better because that exposes us to potential shell injection attacks. I suspect providing the absolute path to the binary would be the better fix, like @gimenete also said.

gimenete:
@fabien0102 @tsg @SferaDev what about this https://github.com/xataio/client-ts/pull/383 ?

PR #382, opened 2022-07-01
Implement --no-input and --force flags. Add flags for optionally reducing interactivity

gimenete:
This PR:

- Wraps `prompts()` in a custom method that checks for interactivity.
- Allows specifying flags for values that can be also interactively requested.
- When interactivity is not possible it uses the `initial` value passed if exists.
- If not, it fails with specific messages.

TODO: see if we can simplify some flags conditionals with this new functionality.

### DIFF
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -57,19 +57,23 @@ export abstract class BaseCommand extends Command {
     description: 'Branch name to use'
   });
 
-  static noInputFlag = Flags.boolean({
-    helpGroup: commonFlagsHelpGroup,
-    description: 'Will not prompt interactively for missing values'
-  });
+  static noInputFlag = {
+    'no-input': Flags.boolean({
+      helpGroup: commonFlagsHelpGroup,
+      description: 'Will not prompt interactively for missing values'
+    })
+  };
 
-  static jsonFlag = Flags.boolean({
-    helpGroup: commonFlagsHelpGroup,
-    description: 'Print the output in JSON format'
-  });
+  static jsonFlag = {
+    json: Flags.boolean({
+      helpGroup: commonFlagsHelpGroup,
+      description: 'Print the output in JSON format'
+    })
+  };
 
   static commonFlags = {
-    json: this.jsonFlag,
-    'no-input': this.noInputFlag
+    ...this.jsonFlag,
+    ...this.noInputFlag
   };
 
   async init() {
@@ -153,7 +157,7 @@ export abstract class BaseCommand extends Command {
         return this.error('No workspaces found, please create one first');
       }
 
-      const { name } = await prompts({
+      const { name } = await this.prompt({
         type: 'text',
         name: 'name',
         message: 'New workspace name'
@@ -167,7 +171,7 @@ export abstract class BaseCommand extends Command {
       return workspace;
     }
 
-    const { workspace } = await prompts({
+    const { workspace } = await this.prompt({
       type: 'select',
       name: 'workspace',
       message: 'Select a workspace',
@@ -197,7 +201,7 @@ export abstract class BaseCommand extends Command {
         choices.splice(0, 0, { title: '<Create a new database>', value: 'create' });
       }
 
-      const { database } = await prompts({
+      const { database } = await this.prompt({
         type: 'select',
         name: 'database',
         message: dbs.length > 0 && options.allowCreate ? 'Select a database or create a new one' : 'Select a database',
@@ -245,7 +249,7 @@ export abstract class BaseCommand extends Command {
         title = branches.length > 0 && options.allowCreate ? 'Select a branch or create a new one' : 'Select a branch'
       } = options;
 
-      const { branch } = await prompts({ type: 'select', name: 'branch', message: title, choices });
+      const { branch } = await this.prompt({ type: 'select', name: 'branch', message: title, choices });
 
       if (!branch) return this.error('No branch selected');
       if (branch === CREATE_CHOICE) {
@@ -264,7 +268,7 @@ export abstract class BaseCommand extends Command {
 
   async createDatabase(workspace: string) {
     const xata = await this.getXataClient();
-    const { name } = await prompts({
+    const { name } = await this.prompt({
       type: 'text',
       name: 'name',
       message: 'New database name',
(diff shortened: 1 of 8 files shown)


cli/src/commands/import/csv.ts
```diff
@@ -19,8 +18,9 @@ export default class ImportCSV extends BaseCommand {
   ];
 
   static flags = {
-    'no-input': this.noInputFlag,
+    ...this.noInputFlag,
     ...this.databaseURLFlag,
+    ...BaseCommand.forceFlag('Update the schema database schema if necessary without asking'),
```
SferaDev, line 23:
```suggestion
    ...BaseCommand.forceFlag('Update the database schema if necessary without asking'),
```

cli/src/base.ts
```diff
@@ -493,4 +506,43 @@ export abstract class BaseCommand extends Command {
...
+
+  async prompt<name extends string>(
+    options: prompts.PromptObject<name>,
+    flagValue?: boolean | string
+  ): Promise<prompts.Answers<name>> {
+    // If there's a flag, use the value of the flag
+    if (flagValue != null) return { [String(options.name)]: flagValue } as prompts.Answers<name>;
+
+    const { flags } = await this.parse(
+      { strict: false, flags: { ...BaseCommand.noInputFlag, force: Flags.boolean() } },
```
SferaDev, line 518:
```suggestion
      { strict: false, flags: { ...BaseCommand.noInputFlag, ...BaseCommand.forceFlag() } },
```
Otherwise -f char won't work, right?

cli/src/base.ts
```diff
@@ -493,4 +506,43 @@ export abstract class BaseCommand extends Command {
...
+    if (!process.stdout.isTTY && process.env.NODE_ENV !== 'test') {
+      reason = 'you are not running it in a TTY';
+    } else if (noInput) {
+      reason = 'the --no-input flag is being used';
+    }
+
+    if (reason) {
+      if (options.type === 'confirm') {
+        this.error(
+          `The current command required interactivity, but ${reason}. Use --force to execute it without asking for confirmations.`
```
SferaDev, line 538: What if I don't want to really force (in terms of overriding) but still answer yes to all confirmation questions?

To me --no-input is an special --force and here it wouldn't fail, it would succeed.

Only if in TTY and --no-input I would fail, and the message would recommend using --no-input instead of --force.

WDYT?

cli/src/commands/auth/login.ts
```diff
@@ -1,4 +1,4 @@
-import prompts from 'prompts';
+import { Flags } from '@oclif/core';
```
SferaDev, line 1:
```suggestion
```
Reviewing online, but this import is not being used in the diff.

cli/src/commands/dbs/delete.ts
```diff
@@ -25,12 +25,15 @@ export default class DatabasesDelete extends BaseCommand {
...
-      type: 'confirm',
-      name: 'confirm',
-      message: `Are you sure you want to delete database ${workspace}/${database}?`,
-      initial: true
-    });
+    const { confirm } = await this.prompt(
+      {
+        type: 'confirm',
+        name: 'confirm',
+        message: `Are you sure you want to delete database ${workspace}/${database}?`,
```
SferaDev, line 32: To be consistent with earlier messages:


```suggestion
        message: `Are you sure you want to delete database ${database} in the ${workspace} workspace?`,
```

cli/src/commands/init/index.ts
```diff
@@ -19,12 +18,19 @@ export default class Init extends BaseCommand {
 
   static flags = {
     ...this.databaseURLFlag,
+    ...BaseCommand.forceFlag('Overwrite existing project configuration'),
+    sdk: Flags.boolean({
+      description: 'Install the TypeScript/JavaScript SDK'
+    }),
+    codegen: Flags.string({
+      description:
+        'Output file where to generate a TypeScript/JavaScript client strictly typed for your database schema'
```
SferaDev, line 27:
```suggestion
        'Output file to generate a TypeScript/JavaScript client with types for your database schema'
```

cli/src/commands/init/index.ts
```diff
@@ -85,34 +95,43 @@ export default class Init extends BaseCommand {
...
-      type: 'text',
-      name: 'output',
-      message: 'Choose where the output file for the code generator',
-      initial: 'src/xata.ts'
-    });
+    const { output } = await this.prompt(
+      {
+        type: 'text',
+        name: 'output',
+        message: 'Choose where the output file for the code generator',
```
SferaDev, line 117:
```suggestion
        message: 'Choose the output file for the code generator',
```

cli/src/commands/workspaces/delete.ts
```diff
@@ -24,12 +24,15 @@ export default class WorkspacesDelete extends BaseCommand {
...
-      type: 'confirm',
-      name: 'confirm',
-      message: `Are you sure you want to delete workspace ${workspace}?`,
-      initial: true
-    });
+    const { confirm } = await this.prompt(
+      {
+        type: 'confirm',
+        name: 'confirm',
+        message: `Are you sure you want to delete workspace ${workspace}?`,
```
SferaDev, line 31:
```suggestion
        message: `Are you sure you want to delete the ${workspace} workspace?`,
```

For consistency with the rest of deletion messages

cli/src/base.ts
SferaDev: --yes sounds great!

SferaDev:
image

Something weird happening here

gimenete:
@SferaDev 🤔 this is my output

image

`xata auth login` does not support `--yes` or `--no-input`, only `--force`.

SferaDev:
The scenario I'm showing is when you already have a token saved. Then at least --force should work, doesn't seem to

gimenete:
> The scenario I'm showing is when you already have a token saved

Likewise. See the first example, it asks for overwritting

gimenete:
With latest published build 
image

SferaDev:
Ah okay. I see, then what's missing for fully non interactive is a way to set up the key. I guess we don't have a flag so it's not saved in bash history. Do we support reading from stdin?

gimenete:
> what's missing for fully non interactive is a way to set up the key

Yup. I don't see a use case right now where a user may want to pipe the value of the API key into the command because it's a value we generate (not something they can provide), so I'd leave it as is.

SferaDev:
Yeah, my only use case is for programmatic usage. Like I'm setting a test that configures xata auth during its execution (what I'm doing with vercel/cloudflare) but I guess in that case I'd use the env variable instead.

SferaDev (approved):

PR #383, opened 2022-07-01
Use which to calculate full path for running commands

gimenete:
On Windows executing `npm` directly doesn't seem to work. Using the full path [seems to work](https://stackoverflow.com/questions/56366448/how-to-fix-error-spawn-npm-enoent-in-windows-10).

### DIFF
--- a/.changeset/cool-ducks-march.md
+++ b/.changeset/cool-ducks-march.md
@@ -1,0 +1,5 @@
+---
+'@xata.io/cli': patch
+---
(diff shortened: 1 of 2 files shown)


SferaDev (approved):
Also works in Windows and better approach
Screenshot 2022-07-01 at 22 02 22

gimenete:
Since it's simpler and safer than the other approach, I'm going to merge.

PR #395, opened 2022-07-04
Fix workflows on forks

SferaDev:
Closes: #392

- Remove `Changeset check`, it's a bit annoying but still helpful? Not sure what to do...
- Disable tests. The majority are integration tests and secrets are not shared. I would love a better approach to be honest.
- Do not attempt publishing a release.

Instead of actions some community open source projects they have a bot to build/test/release and a maintainer pings them via comment in the PR (workaround to avoid this particular issue). Don't think we need that yet, but probably if we end up OSS a lot of things, it could be useful.

Not related to the PR:
- Update an obsolete snapshot test


gimenete (approved):
LGTM. Any contribution can be tested in the release PR so it's fine if we don't test everything in those contributions.

About the changeset, we can also take care of that. Not ideal for getting always a complete changelog but ok for now and less friction for contributions.

PR #398, opened 2022-07-05
Improve staging and REST compatibility

gimenete:
- Generate `databaseURL` values that have into account the profile `api` setting.
- Introduce helper `getCurrentBranchName` function to be used in a couple of places.
- Accept an empty `formatVersion` (what the REST API sends) when parsing a schema.

### DIFF
--- a/cli/src/schema.ts
+++ b/cli/src/schema.ts
@@ -35,7 +35,7 @@ export const tableSchema = z.object({
 export type Table = z.infer<typeof tableSchema>;
 
 export const xataDatabaseSchema = z.object({
-  formatVersion: z.literal('1.0').or(z.literal('')),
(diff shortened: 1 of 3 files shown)


cli/src/schema.ts
```diff
@@ -35,7 +35,7 @@ export const tableSchema = z.object({
 export type Table = z.infer<typeof tableSchema>;
 
 export const xataDatabaseSchema = z.object({
-  formatVersion: z.literal('1.0'),
+  formatVersion: z.literal('1.0').or(z.literal('')),
```
SferaDev, line 38:
```suggestion
  formatVersion: z.literal('1.0').or(z.literal('')).optional(),
```

SferaDev (approved):

PR #409, opened 2022-07-11
Infer codegen types from schema

SferaDev:
## Motivation

- To build the `XataRecord` objects properly we need the schema of the tables
  - Impossible to build links or advanced types like dates without the schema

## Before this PR
- We gathered the schema on a network call and memoized it in the constructor of the client

## Objective
- Remove the extra network call
- Provide the schema with the codegen
- Infer the types from the schema tables

## Breaking changes
- Changed signature of the client constructor (codegen needs to be executed after upgrading)
- Type names are not homogenised (singular/plural and case) anymore (I think this is better to be honest)


gimenete (approved):
Looks great! Awesome job 👏👏👏👏

Super minor thing: probably we can get rid of the `case` package in codegen/package.json

SferaDev:
Will do another PR and try to pass an analyzer that finds for other unused dependencies

Issue #411, opened 2022-07-11
Permission error with xata init --db if there's not a global API key


gimenete:
Reported by Tejas. Seems like the generated API key is not used by `xata init` for the next steps after it is created.

PR #412, opened 2022-07-11
Fix xata init not using newly generated API key

gimenete:
Closes https://github.com/xataio/client-ts/issues/411


richardgill (approved):
LGTM

Issue #407, opened 2022-07-10
Unexpected 403 in client


TejasQ:
I'm experiencing an error from the client and I'm not sure why. 🧐  It worked the first time when bulk importing, but subsequent requests error. 

## Error

<img width="1242" alt="image" src="https://user-images.githubusercontent.com/9947422/178148137-a6cf2732-c372-49bb-a0a6-4e8d732890f2.png">

```
error - FetchError: invalid json response body at https://xata-examples-49fn2g.xata.sh/db/aita:main/tables/posts/data/t3_vvnjbw reason: Unexpected end of JSON input
    at /Users/tejas/Sites/LAB/aita/node_modules/next/dist/compiled/node-fetch/index.js:1:51227
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async fetch$1 (file:///Users/tejas/Sites/LAB/aita/node_modules/@xata.io/client/dist/index.mjs:162:26)
    at async RestRepository.upsertRecordWithID_fn (file:///Users/tejas/Sites/LAB/aita/node_modules/@xata.io/client/dist/index.mjs:1486:20)
    at async RestRepository.createOrUpdate (file:///Users/tejas/Sites/LAB/aita/node_modules/@xata.io/client/dist/index.mjs:1344:22)
    at async Promise.all (index 21)
    at async handler (webpack-internal:///(api)/./pages/api/getData.ts:18:5)
    at async Object.apiResolver (/Users/tejas/Sites/LAB/aita/node_modules/next/dist/server/api-utils/node.js:179:9)
    at async DevServer.runApi (/Users/tejas/Sites/LAB/aita/node_modules/next/dist/server/next-server.js:381:9)
    at async Object.fn (/Users/tejas/Sites/LAB/aita/node_modules/next/dist/server/base-server.js:497:37) {
  status: 403,
  errors: undefined,
  cause: undefined,
  page: '/api/getData'
}
```

## Code

```ts
const handler: NextApiHandler = async (_, res) => {

    // Get and store posts
    const posts = await fetch('https://www.reddit.com/r/amitheasshole.json')
        .then(r => r.json())
        .then(({ data: rootData }) => rootData.children
            .map(({ data }) => ({ id: data.name, title: data.title, content: data.selftext }))
        );

    await client.db.posts.createOrUpdate(posts)
    res.end(posts);
}
```

## Screenshot

[...]

PR #425, opened 2022-07-12
Do not send falsy values to qs

SferaDev:
Context: https://github.com/xataio/client-ts/issues/407#issuecomment-1181620815

### DIFF
--- a/packages/client/src/api/fetcher.ts
+++ b/packages/client/src/api/fetcher.ts
@@ -2,7 +2,13 @@ import { VERSION } from '../version';
 import { FetcherError, PossibleErrors } from './errors';
 
 const resolveUrl = (url: string, queryParams: Record<string, any> = {}, pathParams: Record<string, string> = {}) => {
-  const query = new URLSearchParams(queryParams).toString();
+  // Remove nulls and undefineds from query params
+  const cleanQueryParams = Object.entries(queryParams).reduce((acc, [key, value]) => {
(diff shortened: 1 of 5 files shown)


packages/client/src/api/fetcher.ts
```diff
@@ -2,7 +2,13 @@ import { VERSION } from '../version';
 import { FetcherError, PossibleErrors } from './errors';
 
 const resolveUrl = (url: string, queryParams: Record<string, any> = {}, pathParams: Record<string, string> = {}) => {
-  const query = new URLSearchParams(queryParams).toString();
+  // Remove nulls and undefineds from query params
+  const cleanQueryParams = Object.entries(queryParams).reduce((acc, [key, value]) => {
+    if (value === undefined || value === null) return acc;
```
gimenete, line 7:
```suggestion
    if (value == null) return acc;
```

Nitpick.

gimenete (approved):

packages/client/src/api/fetcher.ts
SferaDev: I tend to be explicit in this particular case and try to avoid coercion. I guess we could have a helper isNullish or something like that to be less verbose everywhere.

gimenete: 👍  sounds good. Yeah, a helper would be nice, to make it less verbose, like you said. But any way works for me.

Issue #427, opened 2022-07-12
Unexpected error about `overrideRecords` 


TejasQ:
I've received this error and I have no idea what to do with it:

<img width="1908" alt="image" src="https://user-images.githubusercontent.com/9947422/178577627-9014ed7a-ad62-4b21-acbd-753d2ba12d2c.png">

Console output:
<img width="1402" alt="image" src="https://user-images.githubusercontent.com/9947422/178578183-96accdd4-bdd7-4760-84c4-7e9d49c173d5.png">

As text:

```
TypeError: (overrideRecords ?? page.records) is not iterable (cannot read property undefined)
```

Logged to the console,

```
{
  speakers: _RecordArray(7) [
    {
      id: 'rec_cb6sg2ihosd005lf78qg',
      image: 'https://github.com/tejasq.png',
      name: 'Xata Team',
      position: '',
      talk: 'Announcing Xata: Now Publicly Available',
      twitter: 'https://twitter.com/tejaskumar_'
    },
    {
      id: 'rec_cb6sgrhh40fcoclfknf0',
      image: 'https://avatars.githubusercontent.com/u/22625602?v=4',
      name: 'Guillaume Bibeau-Laviolette',
      position: 'Solutions Architect, Vercel',
      talk: 'My Favorite Database for Web3',
      twitter: 'https://twitter.com/GuiBibeau'
    },
    {
      id: 'rec_cb6sh7hh40fcoclfknfg',
      image: 'https://github.com/saravieira.png',
      name: 'Sara Vieira',
      position: 'Senior Frontend Engineer, Remote.com',
      talk: 'Creating the 3D Xatafly with react-three-fiber',
      twitter: 'https://twitter.com/nikkitaftw'
    },
    {
      id: 'rec_cb6sho9h40fcoclfkng0',
      image: 'https://github.com/seiflotfy.png',
      name: 'Seif Lotfy',
      position: 'Co-Founder and CTO, Axiom',
      talk: 'A Better Logging Story',
      twitter: 'https://twitter.com/seiflotfy'
    },
    {
      id: 'rec_cb6si3qhosd005lf78r0',
      image: 'https://pbs.twimg.com/profile_images/1493634012220465165/_CUnZ4DC_400x400.jpg',
      name: 'Jenny Truong',
      position: 'COO, Stately.ai',
      talk: 'Modeling Application Logic and Data for Predictable Apps',
      twitter: 'https://twitter.com/jen_ayy_',
      videoUrl: ''
    },
[...]

PR #432, opened 2022-07-13
Fix constructor usage on hydrated `RecordArray`

SferaDev:
Closes: #427

### DIFF
--- a/packages/client/src/schema/pagination.ts
+++ b/packages/client/src/schema/pagination.ts
@@ -1,4 +1,4 @@
-import { isDefined } from '../util/lang';
+import { isDefined, isObject } from '../util/lang';
 import { Query } from './query';
 import { XataRecord } from './record';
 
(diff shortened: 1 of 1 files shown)


packages/client/src/schema/pagination.ts
```diff
@@ -107,10 +107,26 @@ export class RecordArray<Result extends XataRecord> extends Array<Result> {
...
-    super(...(overrideRecords ?? page.records));
+    super(...RecordArray.parseConstructorParams(page, overrideRecords));
     this.#page = page;
   }
 
+  static parseConstructorParams(...args: any[]) {
+    // new <T>(arrayLength: number): T[]
+    if (typeof args[0] === 'number') {
+      return [];
+    }
```
SferaDev, line 118:
```suggestion
    if (args.length === 1 && typeof args[0] === 'number') {
      return new Array(args[0]);
    }
```

gimenete (approved):

Issue #406, opened 2022-07-10
Creating link columns via the CLI fails and I'm not sure why 


TejasQ:
![2022-07-10 15 02 41](https://user-images.githubusercontent.com/9947422/178146043-e356ccb5-8570-4be7-8eb7-1932f6d7d9fa.gif)

I try to create a `link`, but I get this error:

```
Error: type mismatch: type [string] is not valid for field [link]
```

I choose type `link`, not `string` as seen in the reproduction, so I'm not really sure why this error happens. 😅

PR #435, opened 2022-07-13
CLI: fix creating link columns

gimenete:
Closes https://github.com/xataio/client-ts/issues/406

The `answer.values` object is a flat structure and we need `{ link: { table: string } }` instead of `{ link: string }`

### DIFF
--- a/cli/src/commands/schema/edit.ts
+++ b/cli/src/commands/schema/edit.ts
@@ -14,7 +14,7 @@ type Table = Schema['tables'][0];
 type Column = Table['columns'][0];
 
 type EditableColumn = Column & {
-  added?: string;
+  added?: boolean;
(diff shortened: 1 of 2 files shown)


cli/src/commands/schema/edit.ts
```diff
@@ -362,17 +362,25 @@ export default class EditSchema extends BaseCommand {
 
     try {
       const answer = await snippet.run();
+      const { values } = answer;
```
SferaDev, line 365:
```suggestion
      const { values } = await snippet.run();
```

I don't see more usages. Also I think `enquirer.Snippet` does not return `null | undefined`. It's hard to see because we don't have types for the library, any reason?

SferaDev (approved):

PR #437, opened 2022-07-13
Make env var override the xatarc config value

fabien0102:
Little fix about the priority of config values.

This should be `flag -> env var -> .xatarc` to be able to override the value on CI (as example)


gimenete (approved):
LGTM. Can you add a changelog?

gimenete (approved):

PR #440, opened 2022-07-14
Small docs adjustments

TejasQ:
## Proposed Changes

- Add doctoc to provide a nice table of contents automatically for easy navigation
  - Update all ToCs with lint-staged pre-commit
- Adjust some links and typos
- Move `Query` above `Repository` in Usage.md because people usually use the `Query` first, and it's the default path in the Xata SDK playground.

**Preview:** https://docs-git-cli-xata.vercel.app/sdk/getting-started
  ⚠️ The deploy previews on docs.xata.io sometimes don't have syntax highlighting, but `main` does.


packages/client/README.md
```diff
@@ -20,27 +38,25 @@ There are three ways to use the SDK:
...
-});
+const xata = new XataClient();
 ```
 
-The import above will differ if you chose to genreate the code in a different location.
+The import above will differ if you chose to generate the code in a different location.
 
-The `fetch` paramter is required only if your runtime doesn't provide a global `fetch` function. There's also a `databaseURL` argument that by default will contain a URL pointing to your database (e.g. `https://myworkspace-123abc.xata.sh/db/databasename`), it can be specified in the constructor to overwrite that value if for whatever reason you need to connect to a different workspace or database.
+The constructor `XataClient` accepts an object with configuration options, like the `fetch` parameter, which is required only if your runtime doesn't provide a global `fetch` function. There's also a `databaseURL` argument that by default will contain a URL pointing to your database (e.g. `https://myworkspace-123abc.xata.sh/db/databasename`). It can be specified in the constructor to overwrite that value if for whatever reason you need to connect to a different workspace or database.
```
gimenete, line 59:
```suggestion
The `XataClient` constructor accepts an object with configuration options, like the `fetch` parameter, which is required only if your runtime doesn't provide a global `fetch` function. There's also a `databaseURL` argument that by default will contain a URL pointing to your database (e.g. `https://myworkspace-123abc.xata.sh/db/databasename`). It can be specified in the constructor to overwrite that value if for whatever reason you need to connect to a different workspace or database.
```

gimenete (commented):
Doc improvements look good to me.

But I'm not sure about the TOC part. GitHub already provides that in the UI. Maybe it'd be something to do while importing things in the docs site, but leave the markdown files in this repo without it.

image

TejasQ:
Happy to look at adding a table of contents to the docs site, but is there a significant cost to adding the ToCs in place? I see some benefits like:

- it adds links in line, ready to click, with fewer clicks than this button in GitHub
- it's a bit more discoverable than this GitHub button. Full disclosure, I didn't even know it exists 😅 

I'm a bit fuzzy on the downsides/costs of having inline table of contents here, but happy to remove it if it hurts the project for sure!

gimenete:
Besides GitHub already providing the feature to me the main point of friction is that if we edit the markdown in the GitHub UI or someone edits it without having the hook installed, the script that generates the TOC won't run and it may leave the TOC out of sync.

I'd prefer to only have the TOC generated in the docs site.

packages/client/Usage.md
```diff
@@ -1,23 +1,134 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Query
+
+To get a collection of records, you can use the `Query` object. It provides the following methods:
+
+- `getFirst()`: returns the first record in the query results.
+- `getPaginated()`: returns a page of records in the query results.
+- `getAll()`: returns all the records in the query results.
```
atilafassina, line 41: Not discussing the method name here, but there's a threshold for the `getAll` method.

Though conflicting with the method's name, I think it's best to be explicit about it in the docs to avoid confusion and issues for consumers.

```suggestion
- `getAll()`: returns the first 200 records in the query results.
```

packages/client/README.md
```diff
@@ -249,9 +264,7 @@ await client.workspaces.deleteWorkspace(workspace);
 
 ## Deno support
 
-Right now we are still not publishing the client on deno.land or have support for deno in the codegen.
-
-However you can already use it with your preferred node CDN with the following import in the auto-generated `xata.ts` file:
+Right now we are still not publishing the client on deno.land or have support for deno in the codegen. However you can already use it with your preferred Node.js CDN with the following import in the auto-generated `xata.ts` file:
```
atilafassina, line 267:
```suggestion
Right now we are still not publishing the client on [deno.land](https://deno.land) or have support for deno in the codegen. However you can already use it with your preferred Node.js CDN with the following import in the auto-generated `xata.ts` file:
```

TejasQ: Nice touch, thanks!

packages/client/Usage.md
TejasQ: This is actually a really good point to address. I thought get all actually does get all, by getting 200 and then appending the next 200 and the next and the next until all records are available. 

I haven’t yet looked into the implementation so I’m not sure if that happens or if it gets only 200, so thanks! Now I can go dive into the implementation and find out!

SferaDev: That's not what happens 😅

`getAll` gets all the records as I wrote down. There's currently no method that returns 200 also. `getMany` returns 20 which is the default pagination size. And even in that case I wouldn't write the number in the docs, as there's a point of the helper variables that keep track of these "hardcoded" values.

packages/client/README.md
```diff
@@ -249,9 +264,7 @@ await client.workspaces.deleteWorkspace(workspace);
 
 ## Deno support
 
-Right now we are still not publishing the client on deno.land or have support for deno in the codegen.
-
-However you can already use it with your preferred node CDN with the following import in the auto-generated `xata.ts` file:
+Right now we are still not publishing the client on [deno.land](https://deno.land) or have support for deno in the codegen. However you can already use it with your preferred Node.js CDN with the following import in the auto-generated `xata.ts` file:
```
gimenete, line 267: Is it worth putting a link to deno.land if we are saying we don't have anything published there yet?

gimenete (approved):
We can revisit the decision about the TOC later. We'll see if it causes friction.

A suggestion I want to make is that in the future it would be good dicussing changes like that in a separate PR or issue to not block other changes in the same PR. Or leave a message on Slack to give a heads up or discuss it first, before opening a PR.

packages/client/README.md
SferaDev: Should we begin publishing to deno.land?

We need to:

Agree on a name for the deno package. I'm okay with `xata` (they don't have namespaces like npm).

Set-up publishing releases to GitHub automatically.

Configure the webhook.

Can take care of that if we all agree 😎

gimenete: SGTM 👍

SferaDev (approved):

PR #441, opened 2022-07-14
Update test e2e dependencies

gimenete:
Opening this PR because I saw a few [security alerts](https://github.com/xataio/client-ts/security/dependabot) coming from `test/e2e/package.json`. I think maybe this is not covered by the _update dependencies_ workflow file?


SferaDev (approved):
It should be, we are updating with `--deep` so it should check all package.json files in the repo

PR #445, opened 2022-07-15
Improve backend error parsing

SferaDev:


packages/client/src/api/fetcher.ts
```diff
@@ -116,16 +117,17 @@ export async function fetch<
   }
 
   const requestId = response.headers?.get('x-request-id') ?? undefined;
+  const payload = await parsePayload(response);
+  if (response.ok) return payload;
```
gimenete, line 121: Are there any side effects here for returning a string potentially?

gimenete (approved):

packages/client/src/api/fetcher.ts
SferaDev: Good question, our endpoints should always return JSON... Will fix it and throw if it's a string.

SferaDev:
Need to investigate more the scenarios where we get `text`, closing this for now

PR #455, opened 2022-07-18
Allow editing the schema from source

gimenete:
Allow editing the JSON in a temp file using `xata schema edit --source`.

https://user-images.githubusercontent.com/50486/179479016-b3d9fad5-4f3d-45c4-abb5-412e2ac31350.mp4

After recording the video I've added this warning:

image

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@xata.io/cli",
-  "version": "0.5.0",
+  "version": "0.6.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@xata.io/cli",
-      "version": "0.5.0",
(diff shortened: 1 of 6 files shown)


SferaDev (approved):
Tested. Works fine with `nano`, `vim`, etc...

However, didn't work for `code-insiders`, I see some custom logic for `code` not sure if it's related

gimenete:
> However, didn't work for code-insiders, I see some custom logic for code not sure if it's related

I see. For visual editors we'll need to try to get the args that make them stop the process until the editing is done. For the regular vscode is `-w`.

I'll warn the user if the editor is not a terminal editor and the args are unknown.

Thanks for the review!!

gimenete:
I've added support for more editors and I've added more feedback to users if something is not supported.

I've tested with `code-insiders` and it works now! Thanks for the review.

PR #462, opened 2022-07-20
RecordArray constructor params

SferaDev:
- Fix problems with `RecordArray` map in the constructor
- Ensure that the extra methods are lost when doing a `map` operation and it returns a normal array
- Add a helper `toArray` method

### DIFF
--- a/packages/client/src/schema/pagination.ts
+++ b/packages/client/src/schema/pagination.ts
@@ -106,9 +106,10 @@ export function isCursorPaginationOptions(
 export class RecordArray<Result extends XataRecord> extends Array<Result> {
   #page: Paginable<Result, Result>;
 
-  constructor(page: Paginable<any, Result>, overrideRecords?: Result[]) {
-    super(...RecordArray.parseConstructorParams(page, overrideRecords));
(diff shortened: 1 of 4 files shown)


packages/client/src/schema/pagination.ts
```diff
@@ -127,6 +130,14 @@ export class RecordArray<Result extends XataRecord> extends Array<Result> {
     return new Array(...args);
   }
 
+  toArray(): Result[] {
+    return new Array(...this);
+  }
+
+  map<U>(callbackfn: (value: Result, index: number, array: Result[]) => U, thisArg?: any): U[] {
```
gimenete, line 137: Any other method that we should consider overwriting? I don't think so, right?

gimenete (approved):

packages/client/src/schema/pagination.ts
SferaDev: `forEach` doesn't return anything and `reduce` has the return type in the method signature so doesn't need us to change it. I can't think of others to be honest, let's see what users complain in the future 💥

PR #469, opened 2022-07-20
Treat branch names specified with third party env variables as git branches in the resolution algorithm

gimenete:
For example if a Vercel preview environment has `VERCEL_GIT_COMMIT_REF=foo`, we have to resolve `foo` to see if it exists, and if not, use the default branch.

### DIFF
--- a/packages/client/src/util/environment.ts
+++ b/packages/client/src/util/environment.ts
@@ -22,7 +22,7 @@ export function getEnvironment(): Environment {
       return {
         apiKey: process.env.XATA_API_KEY ?? getGlobalApiKey(),
         databaseURL: process.env.XATA_DATABASE_URL ?? getGlobalDatabaseURL(),
-        branch: process.env.XATA_BRANCH,
+        branch: process.env.XATA_BRANCH ?? getGlobalBranch(),
         envBranch: process.env.VERCEL_GIT_COMMIT_REF ?? process.env.CF_PAGES_BRANCH ?? process.env.BRANCH,
         fallbackBranch: process.env.XATA_FALLBACK_BRANCH ?? getGlobalFallbackBranch()
       };
@@ -37,7 +37,7 @@ export function getEnvironment(): Environment {
       return {
         apiKey: Deno.env.get('XATA_API_KEY') ?? getGlobalApiKey(),
         databaseURL: Deno.env.get('XATA_DATABASE_URL') ?? getGlobalDatabaseURL(),
-        branch: Deno.env.get('XATA_BRANCH'),
+        branch: Deno.env.get('XATA_BRANCH') ?? getGlobalBranch(),
         envBranch: Deno.env.get('VERCEL_GIT_COMMIT_REF') ?? Deno.env.get('CF_PAGES_BRANCH') ?? Deno.env.get('BRANCH'),
         fallbackBranch: Deno.env.get('XATA_FALLBACK_BRANCH') ?? getGlobalFallbackBranch()
       };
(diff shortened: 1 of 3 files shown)


packages/client/src/util/environment.ts
```diff
@@ -21,12 +22,9 @@ export function getEnvironment(): Environment {
...
         databaseURL: process.env.XATA_DATABASE_URL ?? getGlobalDatabaseURL(),
-        branch:
-          process.env.XATA_BRANCH ??
-          process.env.VERCEL_GIT_COMMIT_REF ??
-          process.env.CF_PAGES_BRANCH ??
-          process.env.BRANCH ??
-          getGlobalBranch(),
+        branch: process.env.XATA_BRANCH,
+        envBranch:
+          process.env.VERCEL_GIT_COMMIT_REF ?? process.env.CF_PAGES_BRANCH ?? process.env.BRANCH ?? getGlobalBranch(),
```
SferaDev, line 27: `getGlobalBranch` returns global of `XATA_BRANCH`. I believe it should be on the other property not this one.

packages/client/src/util/environment.ts
```diff
@@ -40,8 +38,8 @@ export function getEnvironment(): Environment {
       return {
         apiKey: Deno.env.get('XATA_API_KEY') ?? getGlobalApiKey(),
         databaseURL: Deno.env.get('XATA_DATABASE_URL') ?? getGlobalDatabaseURL(),
-        branch:
-          Deno.env.get('XATA_BRANCH') ??
+        branch: Deno.env.get('XATA_BRANCH'),
```
SferaDev, line 41: Same as in the node.js comment

SferaDev (commented):
Still need to test it, one comment regarding global branch lookup

packages/client/src/util/environment.ts
```diff
@@ -21,12 +22,8 @@ export function getEnvironment(): Environment {
       return {
         apiKey: process.env.XATA_API_KEY ?? getGlobalApiKey(),
         databaseURL: process.env.XATA_DATABASE_URL ?? getGlobalDatabaseURL(),
-        branch:
-          process.env.XATA_BRANCH ??
-          process.env.VERCEL_GIT_COMMIT_REF ??
-          process.env.CF_PAGES_BRANCH ??
-          process.env.BRANCH ??
-          getGlobalBranch(),
+        branch: process.env.XATA_BRANCH,
```
SferaDev, line 25:
```suggestion
        branch: process.env.XATA_BRANCH ?? getGlobalBranch(),
```

packages/client/src/util/environment.ts
```diff
@@ -40,12 +37,8 @@ export function getEnvironment(): Environment {
       return {
         apiKey: Deno.env.get('XATA_API_KEY') ?? getGlobalApiKey(),
         databaseURL: Deno.env.get('XATA_DATABASE_URL') ?? getGlobalDatabaseURL(),
-        branch:
-          Deno.env.get('XATA_BRANCH') ??
-          Deno.env.get('VERCEL_GIT_COMMIT_REF') ??
-          Deno.env.get('CF_PAGES_BRANCH') ??
-          Deno.env.get('BRANCH') ??
-          getGlobalBranch(),
+        branch: Deno.env.get('XATA_BRANCH'),
```
SferaDev, line 40:
```suggestion
        branch: Deno.env.get('XATA_BRANCH') ?? getGlobalBranch(),
```

SferaDev (approved):
LGTM. We still need to fix-up the branch resolution, but this change is definitely needed too

Issue #468, opened 2022-07-20
New workspace name is not accepted by the CLI


TejasQ:
**Describe the bug**

<img width="1200" alt="image" src="https://user-images.githubusercontent.com/9947422/180062525-e3414833-3569-48ff-999b-ae225b607c5a.png">

The CLI asks for a name but expects a slug.

Trying to create the onboarding video revealed this bug.

https://user-images.githubusercontent.com/9947422/180062888-ab8c9370-d2a9-4ef9-a5bc-bb37119d7c12.mp4

**To Reproduce**
Steps to reproduce the behavior.
- If this is a problem with an npm package, please provide a code snippet and a stacktrace if possible.
- If this is a problem with the CLI, please provide the command(s) you are using and the output.

Run `xata init` with no workspaces.

**Expected behavior**

A workspace with my chosen name is created.

**Software version**
 - If this is a problem with an npm package, provide the version you are using in `package.json`.
 - If this is a problem with the CLI, provide the output you get with `xata --version`. 

`@xata.io/cli/0.6.0 darwin-arm64 node-v16.15.1`

**Additional context**

PR #470, opened 2022-07-20
Custom slugify

gimenete:
A custom `slug` function that works with https://github.com/xataio/xata/blob/4397604f6b9b3a33551ebfb25d6cf5b27ec3e4dd/internal/xvalidator/xvalidator.go#L17

We'll probably need to do the same in the frontend.
An alternative would've been that the slug is optional and the backend generates a valid one if not provided.

Closes https://github.com/xataio/client-ts/issues/468

### DIFF
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -9,12 +9,12 @@ import { readFile, writeFile } from 'fs/promises';
 import fetch from 'node-fetch';
 import path from 'path';
 import prompts from 'prompts';
-import slugify from 'slugify';
 import table from 'text-table';
 import which from 'which';
 import { z, ZodError } from 'zod';
 import { createAPIKeyThroughWebUI } from './auth-server.js';
 import { getProfile } from './credentials.js';
+import { slug } from './utils.js';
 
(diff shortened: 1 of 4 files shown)


SferaDev (approved):
LGTM

I think this is a helpful util. What do you think of having it and exporting from the SDK and using it in the CLI and potentially the frontend without repeating the implementation.

Also, debatable but what if instead of doing anything on the CLI, we added this logic to the API call in the SDK. If slug is provided it must be in slug format, if slug is not provided it is auto computed from the name.

gimenete:
@SferaDev those are great ideas. The only thing is that this implementation depends on a third party package. Maybe we can drop that dependency and just do a fully custom implementation with a regex, and then we can move it to the SDK without adding a dependency to it. WDYT?

SferaDev:
Sounds good to me!

Shouldn't be too hard to implement and provides a helpful utility to the SDK (that we and users will likely need)

gimenete:
@SferaDev cool. I'm going to merge this to close the issue it addresses and I've opened https://github.com/xataio/client-ts/issues/473

Issue #465, opened 2022-07-20
`xata codegen` does not use current branch


SferaDev:
**Describe the bug**

When trying to run `xata codegen` on a project properly initialized (has a xatarc file) the generated code is from `main` but not from the git branch I was on.

**To Reproduce**

- Have different schema on a branch different than `main`
- Have a git branch with the same name as the xata branch you want to run codegen on
- Run the codegen command

**Expected behavior**

Codegen runs with the schema from the xata branch you are working on

**Workaround**

You can obviously run `xata codegen -b <branch>` but in a project it should use the context available to guess the branch

**Software version**

- All latest versions from `next` channel

PR #474, opened 2022-07-21
Fix branch resolution when using ESM

gimenete:
When using ESM, `require` is not available and we need to use `import()`.

Closes https://github.com/xataio/client-ts/issues/465

### DIFF
--- a/packages/client/src/util/environment.ts
+++ b/packages/client/src/util/environment.ts
@@ -88,12 +88,19 @@ function getGlobalFallbackBranch(): string | undefined {
 }
 
 export async function getGitBranch(): Promise<string | undefined> {
+  const cmd = ['git', 'branch', '--show-current'];
+
   // Node.js: child_process.execSync
   try {
+    // CJS
     if (typeof require === 'function') {
       const req = require; // Avoid "Detected a Node builtin module import while Node compatibility is disabled" in CloudFlare Workers
-      return req('child_process').execSync('git branch --show-current', { encoding: 'utf-8' }).trim();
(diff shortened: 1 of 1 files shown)


packages/client/src/util/environment.ts
```diff
@@ -88,24 +88,27 @@ function getGlobalFallbackBranch(): string | undefined {
...
   try {
+    // CJS
     if (typeof require === 'function') {
       const req = require; // Avoid "Detected a Node builtin module import while Node compatibility is disabled" in CloudFlare Workers
-      return req('child_process').execSync('git branch --show-current', { encoding: 'utf-8' }).trim();
+      return req('child_process').execSync(cmd.join(' '), { encoding: 'utf-8' }).trim();
     }
+
+    // ESM
+    const { execSync } = await import('child_process');
```
SferaDev, line 102: I wonder if this will trigger warnings on wrangler during its compile time like require did.

Maybe we will need a, although not sure if that will work either...

```
const imp = import
const { execSync } = await imp('child_process');
```

SferaDev: Tested and 🥁 

image

We do need some trick for `import(foo)`

gimenete: Great catch! Should be fixed with this https://github.com/xataio/client-ts/pull/474/commits/22e351ca36914df0e618c697b762529d566776d2

SferaDev (approved):
We should write a blog post: "Hacks to build a multi-platform SDK with a single bundle"

PR #476, opened 2022-07-21
Print URL in the console

gimenete:
~Seems like `open` does not throw/raise an exception if the subprocess fails. We need to check the exitcode of the subprocess, and we need to wait for it to finish to know the exit code.~

~I don't want to always print the URL because it is long and that'd be ugly.~

Finally I've decided to always print a nice, short URL in the console.

### DIFF
--- a/cli/src/auth-server.test.ts
+++ b/cli/src/auth-server.test.ts
@@ -35,7 +35,7 @@ describe('generateURL', () => {
 describe('handler', () => {
   test('405s if the method is not GET', async () => {
     const callback = vi.fn();
-    const httpHandler = handler(privateKey, passphrase, callback);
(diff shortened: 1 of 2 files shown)


SferaDev:
It takes a really long time to me to show (almost 10 seconds) with iTerm2 + SSH + Ubuntu 22.04 + Fish

gimenete:
@SferaDev thanks a lot for testing with that combination! I've changed the approach here https://github.com/xataio/client-ts/pull/476/commits/55079fe37c9ba981d314432bbcb86084b8c2f546 Now it will always print a nice, readable URL. And that's it 😄

SferaDev (approved):

PR #485, opened 2022-07-23
Add support for returning columns

SferaDev:
Implementation:

- Add API changes to support returning columns
- Add overloads for all Create/Read/Update repository operations
- Add column selection to XataRecord
- Remove record cache (as it might be inconsistent with selected columns)


gimenete (approved):
Looks great!! 🥇

Amazing work 👏👏👏👏👏

.github/workflows/build-pr.yml
```diff
@@ -144,6 +144,9 @@ jobs:
       - name: Lint
         run: npm run lint
 
+      - name: Typecheck
+        run: npx tsc
+
```
SferaDev, line 149:
```suggestion
```

SferaDev:
Going to merge this as-is, and since we won't do the release straight away if there's any feedback from the team we can address it before the release

Issue #486, opened 2022-07-23
Review tables schema loading


SferaDev:
It seems to still always request it (per the tests done)

PR #488, opened 2022-07-24
Fix schema tables not being injected

SferaDev:
Closes: #486 

I've added some checks and we indeed now remove the extra call, however seems to be the fastest network call (backend responds from in-memory almost immediately) and it almost does not introduce an improvement.


gimenete (approved):
Even if the improvement is not huge, it is an improvement. Thanks!

PR #496, opened 2022-07-25
Ignore git output in xata client

gimenete:
Closes https://github.com/xataio/client-ts/blob/main/packages/client/src/util/environment.ts#L35

### DIFF
--- a/packages/client/src/util/environment.ts
+++ b/packages/client/src/util/environment.ts
@@ -94,7 +94,7 @@ export async function getGitBranch(): Promise<string | undefined> {
   // Avoid "Detected a Node builtin module import while Node compatibility is disabled" in CloudFlare Workers
   const nodeModule = ['child', 'process'].join('_');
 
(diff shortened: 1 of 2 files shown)


SferaDev (approved):

packages/client/src/util/environment.ts
```diff
@@ -89,29 +89,32 @@ function getGlobalFallbackBranch(): string | undefined {
 
 export async function getGitBranch(): Promise<string | undefined> {
   const cmd = ['git', 'branch', '--show-current'];
+  const fullCmd = cmd.join(' ');
 
   // Avoid "Detected a Node builtin module import while Node compatibility is disabled" in CloudFlare Workers
   const nodeModule = ['child', 'process'].join('_');
 
+  const execOptions = { encoding: 'utf-6', stdio: 'ignore' };
```
gimenete, line 97:
```suggestion
  const execOptions = { encoding: 'utf-8', stdio: 'ignore' };
```

Issue #491, opened 2022-07-25
Provide debug information if an API key is invalid


TejasQ:
**Describe the bug**

Loom: https://www.loom.com/share/4612e5d993014dfc99b4ac797a58a56e

<img width="842" alt="image" src="https://user-images.githubusercontent.com/9947422/180789811-b9a45d94-d223-4634-b2a7-69c07b1ab925.png">

`xata init` always fails with `Invalid API key` when `.env` is present and contains an older API key, despite running `xata auth login` and provide a new API key.

**To Reproduce**
- Install the latest Xata CLI
- Have a `.env` in your current project with a stale API key
- Run `xata auth login`
- Login with either option
- Run `xata init`

**Expected behavior**
`xata auth login`, after successfully logging in, checks if there's an existing stale API key in `.env` and warns that this will still be used, with the choice to override it.

**Software version**
`@xata.io/cli/0.7.0 darwin-arm64 node-v16.15.1`

**Additional Context**

1. This occurred during a demo with another dev advocate
2. ~This blocks creating content like a welcome video, unless we don't show the CLI 😬~

gimenete:
Changing the title because `xata auth login` works as expected.

By design, an API key in an `.env` file will always take precedence over the api key in the credentials file. The reason is that in the near future you may have both but with different scopes, being the one in the `.env` file scoped to the database and thus preferred by operations done in the current database.

So, we'll just provide additional debug information in the error message, to at least let the user know where the API key is obtained from.

TejasQ:
I feel like GitHub should add 💯 to the emoji reactions. If it did, I’d react with that. Thanks!

PR #502, opened 2022-07-26
Provide feedback when the API key is invalid

gimenete:
Closes https://github.com/xataio/client-ts/issues/491

image

### DIFF
--- a/cli/src/commands/auth/status.ts
+++ b/cli/src/commands/auth/status.ts
@@ -1,5 +1,4 @@
 import { BaseCommand } from '../../base.js';
-import { getProfile } from '../../credentials.js';
 
 export default class Status extends BaseCommand {
   static description = 'Check status of the auth settings';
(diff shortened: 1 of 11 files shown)


cli/src/commands/auth/status.ts
```diff
@@ -1,5 +1,4 @@
 import { BaseCommand } from '../../base.js';
-import { getProfile } from '../../credentials.js';
 
 export default class Status extends BaseCommand {
```
SferaDev, line 3: Could we add a log message for `this. apiKeyLocation`?

I have changed the `.env` and the shell env variable to something not valid and it still logs me in 😢 

image

gimenete: `xata auth status` only checks the credentials file, as all subcommands of `xata auth`.

I am thinking on doing maybe a `xata status` command for project-level configuration that tells you a few things:
- If you have a project configured and the path of the configuration file.
- The resolved database URL.
- The resolved branch name.
- Checking the api key in the env file.

But, like I said, it is expected that `xata auth status` only checks the credentials file.

SferaDev (approved):

PR #511, opened 2022-07-29
CLI: project status command

gimenete:
I think it'd be useful to be able to print some information about the current project configuration.

In the future we can add information about the migration status of the current branch.

image

Suggested here for the first time https://github.com/xataio/client-ts/pull/502#discussion_r929829759


richardgill (approved):
LGTM

PR #512, opened 2022-07-30
Change slug function

gimenete:
In order to match the version in the frontend https://github.com/xataio/frontend-next/pull/1144/files

In the near future we'll put it in the SDK probably.

### DIFF
--- a/cli/src/utils.ts
+++ b/cli/src/utils.ts
@@ -2,11 +2,14 @@ export function pluralize(word: string, count: number) {
   return `${word}${count === 1 ? '' : 's'}`;
 }
 
+// Applies to the slug too
+export const MAX_WORKSPACE_NAME_LENGTH = 55;
+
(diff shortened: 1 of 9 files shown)


cli/src/utils.ts
```diff
@@ -1,14 +1,15 @@
-import slugify from 'slugify';
-
 export function pluralize(word: string, count: number) {
   return `${word}${count === 1 ? '' : 's'}`;
 }
 
+// Applies to the slug too
+export const MAX_WORKSPACE_NAME_LENGTH = 55;
+
 // Based on `IsValidIdentifier` from xvalidator.go
```
richardgill, line 8: Non-blocking: Might be worth adding a tactical comment here linking to the very similiar (or identical?) function in frontend-next to protect future-us from accidentally letting them diverge

richardgill (approved):
One non-blocking comment. LGTM

PR #513, opened 2022-07-30
Do not encrypt app info

gimenete:
The goal was to prevent users changing the parameters, but it's not a big deal. Also, we support also not encrypting them because the vscode extension uses web-crypto that does not allow encrypting with a private key.

This will also allow us to simplify the page in the frontend to not use `getServerSideProps`.


richardgill (approved):
LGTM

PR #514, opened 2022-07-30
Ask to introduce workspace/db/branch name to confirm deletions

gimenete:
Like we do in the UI.

### DIFF
--- a/cli/src/commands/branches/delete.ts
+++ b/cli/src/commands/branches/delete.ts
@@ -24,14 +24,15 @@ export default class BranchesDelete extends BaseCommand {
 
     const { confirm } = await this.prompt(
       {
-        type: 'confirm',
+        type: 'text',
         name: 'confirm',
-        message: `Are you sure you want to delete the branch ${database}:${branch} in the ${workspace} workspace?`,
+        message: `Are you sure you want to delete the branch ${database}:${branch} in the ${workspace} workspace? Please type ${branch} to confirm`,
         initial: false
       },
-      flags.force
+      flags.force ? branch : undefined
     );
     if (!confirm) return this.exit(1);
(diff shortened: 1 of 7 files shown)


cli/src/commands/branches/delete.ts
```diff
@@ -24,14 +24,15 @@ export default class BranchesDelete extends BaseCommand {
 
     const { confirm } = await this.prompt(
```
richardgill, line 25: Non-blocking idea:

This pattern appears 3x in this PR, we could factor out a function called `confirmWithPhrase` or something and reuse it.

```
const { confirm } = await this.prompt(
      {
        type: 'text',
        name: 'confirm',
        message: `${message} Please type ${confirmPhrase} to confirm`,
        initial: false
      },
      flags.force ? {confirmPhrase} : undefined
    );
```

cli/src/commands/dbs/delete.ts
```diff
@@ -27,14 +27,15 @@ export default class DatabasesDelete extends BaseCommand {
...
       {
-        type: 'confirm',
+        type: 'text',
         name: 'confirm',
-        message: `Are you sure you want to delete database ${database} in the ${workspace} workspace?`,
+        message: `Are you sure you want to delete database ${database} in the ${workspace} workspace? Please type ${database} to confirm`,
         initial: false
       },
-      flags.force
+      flags.force ? database : undefined
```
richardgill, line 35: Non-blocking:

This line makes me think there might be a way to use `--force` to avoid entering the confirm phrase? 

If so I think adding 3x tests for this is might be a simple copy-paste job?

richardgill (approved):
Left some non-blocking comments.

This is a good idea 👍

Issue #519, opened 2022-08-04
[SDK] Exposing the `branch`


fabien0102:
**Is your feature request related to a problem? Please describe.**
For introspection reason, I would love to have access to the `branch` that `XataClient` use under the hood.

**Describe the solution you'd like**
```ts
const xata = new XataClient();

console.log(xata.branch); // "main"
```

**Describe alternatives you've considered**
I tried to use the `getCurrentBranchName()` method, but since the `databaseURL` is injected as `defaultOptions` in `XataClient` I couldn't use it out-of-box.

SferaDev:
Instead of the proposed:

```ts
xata.branch; // string | undefined
```

Offering:

```ts
const { branch, databaseURL } = await xata.getConfig();
```

Does that work?

PR #524, opened 2022-08-07
Expose branch to runtime

SferaDev:
Closes: https://github.com/xataio/client-ts/issues/519

### DIFF
--- a/packages/client/src/schema/repository.ts
+++ b/packages/client/src/schema/repository.ts
@@ -348,7 +348,7 @@ export class RestRepository<Data extends BaseData, Record extends XataRecord = D
 {
   #table: string;
   #getFetchProps: () => Promise<FetcherExtraProps>;
-  db: SchemaPluginResult<any>;
+  #db: SchemaPluginResult<any>;
   #cache: CacheImpl;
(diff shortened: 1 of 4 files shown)


packages/client/src/schema/repository.ts
```diff
@@ -348,7 +348,7 @@ export class RestRepository<Data extends BaseData, Record extends XataRecord = D
 {
   #table: string;
   #getFetchProps: () => Promise<FetcherExtraProps>;
-  db: SchemaPluginResult<any>;
+  #db: SchemaPluginResult<any>;
```
gimenete, line 351: 🤔 Is this change required for the new `getConfig()` method or is this something you wanted to do anyway?

Should we maek this readonly somehow?

gimenete (approved):
LGTM. Just one question.

Welcome back!! 😄

packages/client/src/schema/repository.ts
SferaDev: Not related, but I noticed it while doing this change.

We don't have the db object in the interface definition. So was available because of JS even if not typed in TS. We shouldn't expose the internals and only public facing API.

Issue #527, opened 2022-08-08
Trying to ensure that users in the same team use the same codegen version


gimenete:
Some ideas:

- [x] Print codegen version in generated file.
- [ ] Ability to have a codegen version in the `engines` field in `package.json`
- [ ] Optionally having codegen as a dependency in the `package.json`
- [ ] Supporting multiple versions of the codegen at the same time. The CLI would select the proper one. Maybe using oclif plugins?

cc @xataio/devrel

SferaDev:
Duplicated

PR #530, opened 2022-08-09
Print codegen version in generated file

gimenete:
This way people know why they get different results from a coworker.

For https://github.com/xataio/client-ts/issues/527

We can remove the `Please do not edit` in the future if we allow to do that at some point.


SferaDev (approved):

SferaDev:
Please add a changeset and we're good to go 😄

PR #542, opened 2022-08-11
Clarify some docs topics per user feedback

TejasQ:
This PR adjust some of the docs issues [Irfan](https://www.notion.so/external-Irfan-b15bbe7f656d43b0b3f4bbe9e22ad0e4) faced. Thanks, Richard!


packages/client/Usage.md
```diff
@@ -38,8 +38,8 @@ To get a collection of records, you can use the `Query` object. It provides the
 
 - `getFirst()`: returns the first record in the query results.
 - `getPaginated()`: returns a page of records in the query results.
-- `getAll()`: returns all the records in the query results.
-- `getMany()`: returns an array of some records in the query results.
+- `getAll()`: returns all the records in the query results. **We recommend avoiding this method and using `getMany` or `getPaginated` instead, as large datasets can cause performance problems in your applications.**
+- `getMany()`: returns 20 results by default (for more defaults, see [helper variables](#helper-variables)), but is customizable via [pagination](#page).
```
SferaDev, line 42:
```suggestion
- `getMany()`: returns an array with a subset of the first results in the query. The default pagination size (20) is used and can be customised by passing a different `{ pagination: { size: number } }` in its options.
```

packages/client/Usage.md
```diff
@@ -38,8 +38,8 @@ To get a collection of records, you can use the `Query` object. It provides the
 
 - `getFirst()`: returns the first record in the query results.
 - `getPaginated()`: returns a page of records in the query results.
-- `getAll()`: returns all the records in the query results.
-- `getMany()`: returns an array of some records in the query results.
+- `getAll()`: returns all the records in the query results. **We recommend avoiding this method and using `getMany` or `getPaginated` instead, as large datasets can cause performance problems in your applications.**
```
SferaDev, line 41:
```suggestion
- `getAll()`: returns all the records in the query results by paginating all the query results. If the query is not filtered and the table is a large dataset, this operation can affect the performance.
```

We should not recommend avoiding this method. There're a lot of cases where the table data is controlled, you don't have pagination and it is okay to use `getAll`. Also if the user properly filters the results, it is the recommended way to get all the records that match the query.

packages/client/Usage.md
TejasQ: This is great. Let's still link to helper variables for more info?

```suggestion
- `getMany()`: returns an array with a subset of the first results in the query. The default [pagination](#page) size (20) is used and can be customised by passing a different `{ pagination: { size: number } }` in its options. To learn more about default sizes, see [helper variables](#helper-variables).
```

SferaDev (dismissed):

packages/client/Usage.md
```diff
@@ -311,7 +311,7 @@ It contains:
 - `meta`: Information about the current page and its cursor.
 
 ```ts
-const page = await xata.db.users.getPaginated();
+const page = await xata.db.users.getPaginated({ pagination: { size: 25, offset: 3 } });
```
richardgill, line 314: This example is helpful for understanding what parameters can be passed in here. 

But there is a trades-off against being the code a user would likely want to copy/paste into their project. Especially the `offset: 3` bit.

I don't have any brilliant suggestion of how to fix 🤔. Maybe we could have a pagination options section which shows how a fullish example of the `getPaginated` params?

SferaDev: There are more options that can be passed to the `query` methods. Why not adding more examples below?

packages/client/Usage.md
```diff
@@ -38,8 +38,8 @@ To get a collection of records, you can use the `Query` object. It provides the
 
 - `getFirst()`: returns the first record in the query results.
 - `getPaginated()`: returns a page of records in the query results.
-- `getAll()`: returns all the records in the query results.
-- `getMany()`: returns an array of some records in the query results.
+- `getAll()`: returns all the records in the query results by paginating all the query results. If the query is not filtered and the table is a large dataset, this operation can affect the performance.
```
richardgill, line 41:
```suggestion
- `getAll()`: returns all the records in the query results by making multiple requests to iterate over all the pages which exist. If the query is not filtered and the table is a large dataset, this operation can affect the performance.
```

packages/client/Usage.md
```diff
@@ -38,8 +38,8 @@ To get a collection of records, you can use the `Query` object. It provides the
...
+- `getMany()`: returns an array with a subset of the first results in the query. The default [pagination](#page) size (20) is used and can be customised by passing a different `{ pagination: { size: number } }` in its options. To learn more about default values, see [helper variables](#helper-variables).
```
SferaDev, line 43:
```suggestion

All these methods allow customising its filters, column selection, column ordering, pagination and even cache TTL. For example:

```ts
// First item sorting by name
const user = await xata.db.users.getFirst({ sort: "name" });

// Get first 50 items but ignore the first one
const users = await xata.db.users.getMany({ pagination: { pageSize: 50, offset: 1 } });

// Get page of 100 items where name contains "foo"
const page = await xata.db.users.getPaginated({ filter: { name: { $contains: "foo" } }, pagination: { pageSize: 100 } });

// Get all admin users and cache the result for 5 minutes
const user = await xata.db.users.filter("role", "admin").getAll({ cache: 5 * 60 * 1000 }); 
```
```

File: packages/client/Usage.md (hunk 3 above)
Comment (SferaDev) [reply packages/client/Usage.md:314]: @richardgill https://github.com/xataio/client-ts/pull/542/files#r943475344

Comment (richardgill) [reply packages/client/Usage.md]: This looks brilliant :)

File: packages/client/Usage.md [hunk 6]
```diff
@@ -38,9 +38,23 @@ To get a collection of records, you can use the `Query` object. It provides the
...
+const user = await xata.db.users.getFirst({ sort: "name" });
+
+// Get first 50 items but ignore the first one
+const users = await xata.db.users.getMany({ pagination: { pageSize: 50, offset: 1 } });
+
+// Get page of 100 items where name contains "foo"
+const page = await xata.db.users.getPaginated({ filter: { name: { $contains: "foo" } }, pagination: { pageSize: 100 } });
+
+// Get all admin users and cache the result for 5 minutes
+const user = await xata.db.users.filter("role", "admin").getAll({ cache: 5 * 60 * 1000 });
```
SferaDev, line 57:
```suggestion
const user = await xata.db.users.filter("role", "admin").getAll({ cache: 5 * 60 * 1000 }); 

// Overwrite values set in a query
const query = xata.db.users.filter("role", "admin").select(["name"]);
const adminUsers = await query.getAll();
const firstAdminUserWithEmail = await query.getFirst({ columns: ["name", "email"] });
```

packages/client/Usage.md
```diff
@@ -38,9 +38,23 @@ To get a collection of records, you can use the `Query` object. It provides the
 
 - `getFirst()`: returns the first record in the query results.
 - `getPaginated()`: returns a page of records in the query results.
-- `getAll()`: returns all the records in the query results.
-- `getMany()`: returns an array of some records in the query results.
+- `getAll()`: returns all the records in the query results by making multiple requests to iterate over all the pages which exist. If the query is not filtered and the table is a large dataset, this operation can affect the performance.
+- `getMany()`: returns an array with a subset of the first results in the query. The default [pagination](#page) size (20) is used and can be customised by passing a different `{ pagination: { size: number } }` in its options. To learn more about default values, see [helper variables](#helper-variables).
 
+All these methods allow customising its filters, column selection, column ordering, pagination and even cache TTL. For example:
```
SferaDev, line 44:
```suggestion
All these methods allow customising its filters, column selection, column ordering, pagination or cache TTL. For example:
```

packages/client/Usage.md
```diff
@@ -38,9 +38,28 @@ To get a collection of records, you can use the `Query` object. It provides the
...
+// Get page of 100 items where name contains "foo"
+const page = await xata.db.users.getPaginated({ filter: { name: { $contains: "foo" } }, pagination: { pageSize: 100 } });
+
+// Get all admin users and cache the result for 5 minutes
+const user = await xata.db.users.filter("role", "admin").getAll({ cache: 5 * 60 * 1000 }); 
+
+// Overwrite values set in a query
+const query = xata.db.users.filter("role", "admin").select(["name"]);
+const adminUsers = await query.getAll();
+const firstAdminUserWithEmail = await query.getFirst({ columns: ["name", "email"] });
```
SferaDev, line 62:
```suggestion
const firstAdminUserWithEmail = await query.getFirst({ columns: ["name", "email"] });
```
```

SferaDev (approved):

Issue #540, opened 2022-08-11
No `fetch` error should provide instructions on how to resolve


richardgill:
Current error message: `The \`fetch\` option passed to the Xata client is resolving to a falsy value and may not be correctly imported.` [code](https://github.com/xataio/client-ts/blob/a5f34908ad5b4560209edf323f164ef324aaccbd/packages/client/src/util/fetch.ts#L9)

--- 
It would be good to link to docs explaining how to resolve this. 

The fix is roughly:

Couldn't find `fetch`.

If you're using node <18 we recommend you install node-fetch `npm install --save node-fetch` or `yarn add node-fetch`

And pass fetch to the XataClient constructor:

```
import fetch from "node-fetch";

new XataClient({ 
  //...other options
  fetch 
})
```

PR #548, opened 2022-08-11
Client: copy changes

gimenete:
Closes https://github.com/xataio/client-ts/issues/540

This function is used in a few places, not just in `XataClient`, so I tried to list all possible constructors or functions the user may be using.

WDYT @richardgill ?


packages/client/src/util/fetch.ts
```diff
@@ -6,7 +6,7 @@ export function getFetchImplementation(userFetch?: FetchImpl) {
   if (!fetchImpl) {
     /** @todo add a link after docs exist */
     throw new Error(
-      `The \`fetch\` option passed to the Xata client is resolving to a falsy value and may not be correctly imported.`
+      `Couldn't find \`fetch\`. Install a fetch implementation such as \`node-fetch\` and pass it explicitly to the XataClient or XataApiClient constructors or the getCurrentBranchXXX() functions`
```
SferaDev, line 9:
```suggestion
      `Couldn't find \`fetch\`. Install a fetch implementation and pass it explicitly.`
```

And probably a link to a docs page.

SferaDev (approved):

SferaDev:
Once the docs page is created, we will remove the mention of `node-fetch` and add the link

Issue #538, opened 2022-08-10
SDK: Add filters / helpers with long names


richardgill:
We have `gt`, `gte`, `lt` etc.

We could **also** have: `greaterThan`, `greaterThanEqual`, `lessThan` etc.

Issue #536, opened 2022-08-10
SDK cannot filter using `gt` + datetimes


richardgill:
**Describe the bug**
![image](https://user-images.githubusercontent.com/6640874/183960536-b7ffe792-baf5-4563-b085-a460db8ba80e.png)

PR #551, opened 2022-08-12
Filters and dates

SferaDev:
Closes: #536
Closes: #538

### DIFF
--- a/packages/client/src/schema/operators.ts
+++ b/packages/client/src/schema/operators.ts
@@ -14,6 +14,11 @@ import { SelectableColumn } from './selection';
 export const gt = <T extends ComparableType>(value: T): ComparableTypeFilter<T> => ({ $gt: value });
 
 /**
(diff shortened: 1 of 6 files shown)


packages/client/src/schema/operators.ts
```diff
@@ -13,21 +13,51 @@ import { SelectableColumn } from './selection';
  */
 export const gt = <T extends ComparableType>(value: T): ComparableTypeFilter<T> => ({ $gt: value });
 
+/**
+ * Operator to restrict results to only values that are greater than the given value.
+ */
+export const greaterThan = <T extends ComparableType>(value: T): ComparableTypeFilter<T> => ({ $gt: value });
```
gimenete, line 19:
```suggestion
export const greaterThan = gt;
```

What about doing this with aliases?

SferaDev: Sure!

gimenete (approved):
✨

PR #554, opened 2022-08-12
Add edge caching support

SferaDev:
New plugin for caching using KV.

This is more to review the approach for cache plugins rather than the actual code.

CI Fails until I manually create the package in npm.


gimenete (approved):
Cool!!

Do you think it's worth adding some tests?

SferaDev:
> Do you think it's worth adding some tests?

Yes, although since I might change the strategies I'd wait until I have a more clear view of what we offer/support. This is a candidate to have end to end tests and introspect the cache with the API.

Issue #559, opened 2022-08-15
Add methods that throw


fabien0102:
**Describe the bug**
With the following code (simple expressjs app)
```ts
// Delete climber
  app.delete("/climbers/:id", async (req, res) => {
    const id = req.params.id;

    try {
      console.log("try");
      await xata.db.climbers.delete(id);
      console.log("success", id);

      res.status(204).send();
    } catch {
      res.status(400).json({
        message: "Can't delete this record",
      });
    }
  });
```

I was expecting a `400` status if I send `"yolo"` as `id` but instead I have a `204` and this in my console:

![image](https://user-images.githubusercontent.com/1761469/184653126-23ee4f77-c797-4c15-8a39-817a3bd57a14.png)

**Expected behavior**
I was expecting to throw an exception since the `id` is not valid.

**Software version**

`"@xata.io/client": "0.16.1"`

SferaDev:
We don't throw on any error, when there's a situation not possible we return null. We have discussed in the past adding `actionOrThrow` methods: https://xata-hq.slack.com/archives/C032EKCBUGM/p1657121418430689?thread_ts=1657121418.430689&cid=C032EKCBUGM

Will keep this issue to implement those.

fabien0102:
![image](https://user-images.githubusercontent.com/1761469/185080029-7227ecd4-4738-4af0-866f-e528c6529f6a.png)

I'm sorry, but from the documentation, this should already throws, nop? 🤔

PR #562, opened 2022-08-17
Add methods that can throw

SferaDev:
Closes: #559


test/integration/read.test.ts
```diff
@@ -75,13 +105,30 @@ describe('record read', () => {
...
+    expect(result).toHaveLength(items.length);
+    expect(result).toEqual(items.map(() => null));
+  });
+
   test('read multiple with falsy values, throws', async () => {
+    const items = [null, undefined, false, 0, ''];
+
     // @ts-ignore
-    expect(client.db.teams.read([null, undefined, false, 0, ''])).rejects.toThrow();
+    await expect(client.db.teams.readOrThrow(items)).rejects.toThrow();
```
SferaDev, line 132:
```suggestion
   expect(client.db.teams.readOrThrow(items)).rejects.toThrow();
```

test/integration/read.test.ts
```diff
@@ -75,13 +105,30 @@ describe('record read', () => {
...
+    expect(result).toHaveLength(items.length);
+    expect(result).toEqual(items.map(() => null));
+  });
+
   test('read multiple with falsy values, throws', async () => {
+    const items = [null, undefined, false, 0, ''];
+
     // @ts-ignore
-    expect(client.db.teams.read([null, undefined, false, 0, ''])).rejects.toThrow();
+   expect(client.db.teams.readOrThrow(items)).rejects.toThrow();
```
SferaDev, line 132:
```suggestion
    expect(client.db.teams.readOrThrow(items)).rejects.toThrow();
```

SferaDev:
After a meeting with @tsg we decided:

- `Record` methods like `team.read`, `team.update` or `team.delete`, throw if record has been deleted.
- `getFirst`, `read`, `update`, `delete` are safe, and we create their `orThrows` variants
- `create` is safe but the return type if `id` is not passed doesn't include `null`
  - The safe operation without `null` on `create` is `createOrUpdate`

packages/client/src/schema/repository.ts
```diff
@@ -1039,11 +1440,17 @@ export const initObject = <T>(
...
+        } else {
+          result[column.name] = null;
         }
 
         break;
       }
       default:
+        result[column.name] = value ?? null;
+        if (column.notNull === true && value === null) {
+          console.error(`Parse error, column ${column.name} is non nullable and value resolves null`);
```
SferaDev, line 1452: I hope this doesn't happen, I considered throwing but wasn't sure. Made it safe in the end...

SferaDev:
With a bit of luck this is the last breaking change that I think we need to introduce with our current features.

packages/client/src/schema/repository.ts
gimenete: 👍 to make it safe.

I think at some point we'll need a "verbose/debug" option or an optional "log" function or something like that to decide how to print these messages. But of course, out of the scope of this PR. This is a very edge case and it's fine to leave it here for now.

gimenete (approved):
Fantastic job 👏 👏 👏  🥇

Issue #537, opened 2022-08-10
Clearer SDK types in vscode


richardgill:
SDK types can be very hard to understand (both params and responses).

The typing we have feels right, but it would be great if vscode showed the types more concretely / simply.

Here is an example of where it got complicated:

![image](https://user-images.githubusercontent.com/6640874/183961063-da3d6a86-4977-4c3c-9b1b-e8b5488d16b5.png)

PR #564, opened 2022-08-18
Make types more clear

SferaDev:
Improves: #537


richardgill (approved):
This LGTM

I did some thorough testing and it's an improvement, but we have follow up ideas to fully solve  #537

PR #573, opened 2022-08-23
Re-organize test files

SferaDev:
- Move test setup in a common file
- Allow running tests from staging and locally
- Add tracing to tests

Note: This will be used as the foundation for the first E2E tests in the backend

New env variables only for testing:

```
XATA_API_PROVIDER=staging
XATA_API_PROVIDER=http://localhost:6000,http://{workspaceId}.localhost:6000
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
OTEL_RESOURCE_ATTRIBUTES=key=value,foo=bar
```


test/integration/branches.test.ts
```diff
@@ -1,34 +1,50 @@
...
 import { XataClient } from '../../packages/codegen/example/xata';
+import { setUpTestEnvironment, TestEnvironmentResult } from '../utils/setup';
+
+let api: XataApiClient;
+let workspace: string;
+let database: string;
+let hooks: TestEnvironmentResult['hooks'];
+let fetch: TestEnvironmentResult['clientOptions']['fetch'];
+
+let getBranchOptions: {
```
richardgill, line 14: Nit (non-blocking): I might be misunderstanding. But when I was reading this code I thought this was a function because it starts with get.

test/utils/setup.ts
```diff
@@ -0,0 +1,150 @@
...
+  };
+};
+
+export async function setUpTestEnvironment(
+  prefix: string,
+  { cache }: EnvironmentOptions = {}
+): Promise<TestEnvironmentResult> {
+  const { trace, tracer } = await setupTracing();
+
+  const id = Math.round(Math.random() * 100000);
```
richardgill, line 61: I'm not sure, but do we want to use a uuid for this?

richardgill (approved):
Left a couple of comments, but LGTM

test/utils/setup.ts
SferaDev: Yeah, we faced some random 422 already. I was thinking of timestamp instead

PR #576, opened 2022-08-24
Allow empty filters

SferaDev:
Closes: #565
Closes: #566


test/integration/query.test.ts
```diff
@@ -204,6 +204,16 @@ describe('integration tests', () => {
     expect(teams).toHaveLength(0);
   });
 
+  test('filter is empty', async () => {
```
richardgill, line 207: I saw one of the issues was for `.filter({})` should we add a quick test for that too?

richardgill (approved):
LGTM, added 1 comment

test/integration/query.test.ts
SferaDev: Also included in that test! It's actually kind of empty (we don't send any filters to the API)

PR #588, opened 2022-08-26
Fix CLI on Windows: fork Node.js process to pass custom NODE_OPTIONS

gimenete:
`#!/usr/bin/env -S ...` doesn't work on Windows, and we need a way to pass custom `NODE_OPTIONS`.

With this PR we create a fork of the Node.js process and we pass the `NODE_OPTIONS` that we want, keeping a simple shebang that should work on any platform.

Things I've tested:
- running directly works
- interactivity works
- running with npx works
- args work
- other env variables passed to the main process are passed to the child process as well

### DIFF
--- a/cli/bin/run.js
+++ b/cli/bin/run.js
@@ -7,8 +7,11 @@ import { fileURLToPath } from 'url';
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
-fork(join(__dirname, 'run-oclif.js'), {
+const args = process.argv.slice(2);
+
+fork(join(__dirname, 'run-oclif.js'), args, {
   env: {
+    ...process.env,
     NODE_OPTIONS: '--experimental-vm-modules'
   }
 });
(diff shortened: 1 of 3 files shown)


fabien0102:
image

Working for me 💯

cli/bin/run.js
```diff
@@ -1,6 +1,17 @@
...
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const args = process.argv.slice(2);
+
+fork(join(__dirname, 'run-oclif.js'), args, {
+  env: {
+    ...process.env,
+    NODE_OPTIONS: '--experimental-vm-modules'
```
gimenete, line 15: Maybe we can move this up, so we can override `NODE_OPTIONS` when testing?

SferaDev (approved):

cli/bin/run.js
SferaDev, line 15:
```suggestion
    ...process.env,
    NODE_OPTIONS: `${process.env.NODE_OPTIONS ?? ""} --experimental-vm-modules`
```

PR #598, opened 2022-08-30
Add patch database metadata

SferaDev:

### DIFF
--- a/packages/client/src/api/client.ts
+++ b/packages/client/src/api/client.ts
@@ -274,6 +274,18 @@ class DatabaseApi {
     });
   }
 
+  public patchDatabaseMetadata(
+    workspace: Schemas.WorkspaceID,
+    dbName: Schemas.DBName,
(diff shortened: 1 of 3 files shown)


packages/client/src/api/client.ts
```diff
@@ -274,6 +274,18 @@ class DatabaseApi {
     });
   }
 
+  public patchDatabaseMetadata(
```
gimenete, line 277: What about `updateDatabaseMetadata`? I know the openapi spec id uses `patch` in its name, but... we are not doing `patch()` when updating records for example.

gimenete (approved):
Looks good. Just a comment on the naming.

packages/client/src/api/client.ts
SferaDev: Agreed I even had a comment on the openapi side but I was ignored 😅

PR #604, opened 2022-08-31
[tracing] Propagate TraceID to the backend. Create a new setupEnv span. Fix span hierarchy

sgirones:
This PR tackles three main issues:
* *Propagate TraceID to the backend:* this change enables Datadog APM to link the SDK traces with the backend traces and logs
* *Use a dedicated `span` for the setup actions:* this change traces the setup actions performed by every `test suite` and links the `setup span` to the `suite span`

TODO:
- [ ] *Fix span hierarchy:* Actions performed inside each `test case` are linked to the `suite span` instead of the `test span`. The test span created in the `beforeEach` hook needs to be activated before the test code runs.

### DIFF
--- a/packages/client/src/schema/tracing.ts
+++ b/packages/client/src/schema/tracing.ts
@@ -2,17 +2,26 @@ export type AttributeDictionary = Record<string, string | number | boolean | und
 
 export type TraceFunction = <T>(
   name: string,
-  fn: (options: { setAttributes: (attrs: AttributeDictionary) => void }) => T,
+  fn: (options: {
+    setAttributes: (attrs: AttributeDictionary) => void;
+    propagateTrace: (headers: Record<string, any>) => void;
+  }) => T,
   options?: AttributeDictionary
 ) => Promise<T>;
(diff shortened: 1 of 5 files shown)


packages/client/src/schema/tracing.ts
```diff
@@ -1,19 +1,32 @@
 export type AttributeDictionary = Record<string, string | number | boolean | undefined>;
 
+export type SetAttributesFn = (
```
sgirones, line 3: this one is unused it seems

SferaDev: Yeah these are publicly exported types, so doesn't matter if we don't use them internally.

sgirones:
Looking good! The hierarchy of spans is maintained till the backend 👌

One downside with the final refactor is that we lost the parent Span, and now we have a bunch of top-level spans that belong to the same suite but we don't have a way to point to "all of them".

SferaDev:
I will look at it! Thanks @sgirones

PR #610, opened 2022-09-05
Add support for not null and unique in interactive schema editor

gimenete:
- Add support for `UNIQUE`.
- Add support for `NOT NULL` behind a feature flag. Feature flags can be enabled using an env variable. Example for this PR: `XATA_CLI_FLAGS=notnull xata schema edit`


.changeset/grumpy-jeans-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xata.io/cli': patch
+---
+
+dd support for not null and unique in interactive schema editor
```
SferaDev, line 5:
```suggestion
Add support for not null and unique in interactive schema editor
```

cli/src/commands/schema/edit.ts
```diff
@@ -298,16 +299,22 @@ Beware that this can lead to ${chalk.bold(
...
+      getType(),
+      column.unique ? chalk.gray.italic('unique') : '',
+      column.notNull ? chalk.gray.italic('not null') : ''
+    ]
+      .filter(Boolean)
+      .join(' ');
     if (table.deleted || column.deleted || linkedTable?.deleted)
-      return `- ${chalk.red.strikethrough(column.name)} ${type}`;
-    if (table.added || column.added) return `- ${chalk.green(column.name)} ${type}`;
+      return `- ${chalk.red.strikethrough(column.name)} (${metadata})`;
```
SferaDev, line 313: If `metadata` is "" it will print:

```
- ColumnName ()
```

cli/src/commands/schema/edit.ts
```diff
@@ -623,3 +678,44 @@ Beware that this can lead to ${chalk.bold(
...
+  if (['false', 'f', '0', 'n', 'no'].includes(val)) return false;
+  return null;
+}
+
+function validateOptionalBoolean(value?: string) {
+  const bool = parseBoolean(value);
+  if (bool === null) {
+    return 'Please enter a boolean value (e.g. yes, no, true, false) or leave it empty';
+  }
+  return true;
```
SferaDev, line 695: Shouldn't it be:

```suggestion
  return bool;
```

cli/src/feature-flags.ts
```diff
@@ -0,0 +1,8 @@
+const flags = (process.env.XATA_CLI_FLAGS || '')
```
SferaDev, line 1: I believe we should have this in `BaseCommand` and allow them be set from flags in a global way too or env variable (WDYT?)

cli/src/commands/schema/edit.ts
gimenete: But the metadata always contains at least the type, so it won't be an empty string.

cli/src/commands/schema/edit.ts
gimenete: yeah... if you return a string it will be an error message that will be printed to the user. If the value is valid you just return `true` 🤷

SferaDev (approved):

cli/src/commands/schema/edit.ts
SferaDev: True!

cli/src/feature-flags.ts
gimenete: I thought that for things like this we don't want them to expose it with flags, so that's why I didn't put it as ocliff flags. And since it's something that can be isolated and the BaseCommand class is pretty big already, I just put it in a different file.

gimenete: We can revisit it anyway in the future if we add more flags like this.

SferaDev: I already had one for workers and I added as a hidden flag in the two commands affected.

PR #611, opened 2022-09-05
Use multiple env files

gimenete:
- Use `.env.local` if it exists for storing env vars during `xata init`.
- Load `.env.local` and `.env` always in the CLI.

### DIFF
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -104,9 +104,10 @@ export abstract class BaseCommand extends Command {
   }
 
   loadEnvFile(path: string) {
+    const apiKey = process.env.XATA_API_KEY;
     let env = dotenv.config({ path });
     env = dotenvExpand.expand(env);
-    if (!this.apiKeyDotenvLocation && env.parsed?.['XATA_API_KEY']) {
+    if (!apiKey && env.parsed?.['XATA_API_KEY']) {
       this.apiKeyLocation = 'dotenv';
       this.apiKeyDotenvLocation = path;
     }
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -42,8 +42,8 @@ export abstract class BaseCommand extends Command {
   projectConfig?: ProjectConfig;
   projectConfigLocation?: string;
 
-  dotenvLocation = '.env';
   apiKeyLocation?: APIKeyLocation;
+  apiKeyDotenvLocation = '';
 
   #xataClient?: XataApiClient;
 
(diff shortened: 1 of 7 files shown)


cli/src/base.ts
```diff
@@ -100,10 +100,18 @@ export abstract class BaseCommand extends Command {
...
+      this.apiKeyDotenvLocation = path;
+    }
+  }
+
   async init() {
     if (process.env.XATA_API_KEY) this.apiKeyLocation = 'shell';
-    const env = dotenv.config({ path: this.dotenvLocation });
-    if (env.parsed?.['XATA_API_KEY']) this.apiKeyLocation = 'dotenv';
+    this.loadEnvFile('.env.local');
+    this.loadEnvFile('.env');
```
atilafassina, line 114: I don't think this is enough.

As I mentioned in the thread, it's best that `.env.*` takes priority instead of just local.

If you check the Next.js docs, it's what they do too. They even recommend different patterns such as `.env.development` and `.env.production`

If the user has multiple, they need to configure, that's ok. But if they have only `.env.development` will not work with this implementation, and it's a decent use-case.

atilafassina (changes requested):

cli/src/base.ts
gimenete: @atilafassina from a CLI perspective there's no "environment". You are working locally.

atilafassina (approved):

cli/src/base.ts
SferaDev, line 114: Doesn't this overwrite the location on subsequent calls?

So, if user has `.env.local` and `.env`, it will pass all truthy conditional branches and write twice the variable, first with local and then with the normal one.

Similarly to the write below, I would pass an array and early return once first found.

SferaDev:
Also I would mock `fs.readFile` and add a test for this scenario

cli/src/base.ts
gimenete: Yes, it overwrites the location, but that's intentional. If you have a `XATA_API_KEY` in both `.env.local` and `.env`, then the one in `.env` overwrites the one in `.env.local`.

I can do also an array, but the behavior won't change.

SferaDev: Shouldn't it be the opposite?

If I have it in my local one, that's the one that should be used, not the one in normal dotenv 🤔

gimenete: 🤔 I was following the next.js docs. After re-reading them I think you are right 👍

gimenete: For the future: I don't think we want to implement this. Next.js uses `NODE_ENV` to load those files, and `NODE_ENV` typically is used inside a Node.js process. In our case you are invoking the CLI from a shell, not from a Node.js process. Besides, you are either running the CLI locally or from CI. If you are using the CLI locally, then `local` is assumed. And if you are running the CLI from CI then you can just set environmental variables in the shell before running the CLI.

SferaDev (approved):
LGTM. It parses correctly in order the different dotenv files.

I found a problem though, where we use `getProfile` (like `xata auth status` instead of loading it from env it gets the credentials one. I think we should refactor that part a little bit and also take into account the credentials one as last resort (if user is not in a project)

PR #615, opened 2022-09-06
Make `getMany` return more than max size items

SferaDev:


packages/client/src/schema/query.ts
```diff
@@ -332,14 +333,28 @@ export class Query<Record extends XataRecord, Result extends XataRecord = Record
...
+    let more = page.hasNextPage();
+    const results = [...page.records];
+
+    while (more && results.length < size) {
+      page = await page.nextPage();
+      more = page.hasNextPage();
+      results.push(...page.records);
+    }
+
+    if (page.hasNextPage()) {
```
SferaDev, line 350:
```suggestion
    let page = await this.getPaginated({ ...rest, pagination: { size: batchSize, offset } });
    const results = [...page.records];

    while (page.hasNextPage() && results.length < size) {
      page = await page.nextPage();
      results.push(...page.records);
    }

    if (page.hasNextPage()) {
```

tsg:
Hmm, I don’t know, this makes `getMany` do multiple round trips, and I’m afraid that is going to be surprising to people.

I saw the ticket from @TejasQ , but I thought your other PR that adds limit to `getAll` is enough.

packages/client/src/schema/query.ts
```diff
@@ -332,14 +333,26 @@ export class Query<Record extends XataRecord, Result extends XataRecord = Record
...
-    if (page.hasNextPage() && options.pagination?.size === undefined) {
+    let page = await this.getPaginated({ ...rest, pagination: { size: batchSize, offset } });
+    const results = [...page.records];
+
+    while (page.hasNextPage() && results.length < size) {
+      page = await page.nextPage();
+      results.push(...page.records);
+    }
+
+    if (page.hasNextPage()) {
```
SferaDev, line 348:
```suggestion
    if (page.hasNextPage() && options.pagination?.size === undefined) {
```

SferaDev:
@tsg It will only make multiple roundtrips if the user specifies a `size` bigger than the maximum pagination size (200), otherwise it will make just one as expected.

My idea is that `getPaginated` behaves exactly as the API, but our special methods like `getMany` or `getAll` don't necessarily match the API parameters.

We can describe the behaviour in the docs, although I don't think TypeScript developers building queries will be thinking of the number of roundtrips something is doing if they are kept to the minimum to achieve what they have asked for.

PR #661, opened 2022-09-28
Add headers for network call correlations

SferaDev:
Add `ClientID` on constructor build
Add `SessionID` on operation call


packages/client/src/util/uuid.ts
```diff
@@ -0,0 +1,7 @@
+export function generateUUID() {
+  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+    const r = (Math.random() * 16) | 0,
+      v = c == 'x' ? r : (r & 0x3) | 0x8;
+    return v.toString(16);
+  });
+}
```
SferaDev, line 7: Not 100% happy with this, we shouldn't rely on libraries or packages though...

gimenete: I think it's good enough 😄

gimenete (approved):

PR #662, opened 2022-09-28
Add aggregations/summaries support

SferaDev:
- Add `aggregate` at the `Repository` level
- Add `summarize` at the `Query` level

Both operations are type-safe and:

1. Infer the return types from the `aggregations` and `summaries`
2. Infer the rest of options from the input expressions


packages/codegen/example/schema.json
```diff
@@ -30,11 +63,46 @@
...
+        },
         {
           "name": "full_name",
           "type": "string",
           "notNull": true
         },
+        {
+          "name": "email",
+          "type": "email"
+        },
```
SferaDev, line 97:
```suggestion
```

test/integration/aggregate.test.ts
```diff
@@ -0,0 +1,313 @@
...
+    expect(result.aggs.byIndex.values?.[2].$key).toBe(10);
+    expect(result.aggs.byIndex.values?.[2].$count).toBe(1);
+    expect(result.aggs.byIndex.values?.[2].avgRating).toBe(5.3);
+  });
+});
+
+async function waitForSearchIndexing(): Promise<void> {
+  const { teams = [] } = await xata.search.byTable('longer');
+  if (teams.length === 0) {
+    await new Promise((resolve) => setTimeout(resolve, 2000));
```
gimenete, line 310: Maybe we could make this timeout configurable via env variable.

test/integration/aggregate.test.ts
```diff
@@ -0,0 +1,313 @@
...
+          calendarInterval: 'week',
+          aggs: {
+            total: { count: '*' },
+            unique: { uniqueCount: { column: 'name' } }
+          }
+        }
+      }
+    });
+
+    expect(result.aggs.byDate.values).toHaveLength(3);
```
gimenete, line 193: Nitpick. In these tests we could extract this to a variable to make the other lines a bit more DRY.

```ts
    const values = result.aggs.byDate.values;
```

gimenete (approved):
Looks great. Just a couple of minor comments.

test/integration/aggregate.test.ts
SferaDev: I'll clean the tests a bit in a follow up PR.

Being honest the test have mainly been written by Copilot translating them from the backend ones in go 😍🙃

I just gave it as input the type definition of the method and the input/output values of all the test cases, and it correctly generated function calls with the correct expects.

I love the future

test/integration/aggregate.test.ts
SferaDev: It's not a timeout per se, it's a back off time in the while loop. For timeout we use the test suite one which is 2 minutes. We could make that one configurable from outside (I'll leave it for a future PR too).

PR #676, opened 2022-10-10
Add E2E test that ensures backwards compat with TS Versions

SferaDev:

### DIFF
--- a/.github/workflows/tsc.yml
+++ b/.github/workflows/tsc.yml
@@ -1,0 +1,37 @@
+name: TypeScript E2E Test
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  tsc:
+    name: TypeScript Version Check
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' }}
+
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/tsc.yml
```diff
@@ -0,0 +1,39 @@
+name: TypeScript E2E Test
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  tsc:
+    name: TypeScript Version Check
+    runs-on: ubuntu-latest
```
SferaDev, line 10: Add conditional to only run on changeset releases branches

.github/workflows/tsc.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        target: ['next', 'beta', 'latest', '~4.7']
```
SferaDev, line 15: We will manually need to add a new entry whenever TS releases a new version

.github/workflows/tsc.yml
SferaDev, line 10:
```suggestion
    runs-on: ubuntu-latest
    if: ${{ github.event_name == 'workflow_dispatch' || github.head_ref == 'changeset-release/main' }}
```

gimenete (approved):

PR #689, opened 2022-10-13
Update OpenAPI spec to new tags

SferaDev:


SferaDev:
Will do together with Multi region

PR #704, opened 2022-10-19
Send extra properties in object creation or update

SferaDev:


packages/client/src/util/fetch.ts
```diff
@@ -11,3 +24,70 @@ export function getFetchImplementation(userFetch?: FetchImpl) {
...
+
+    return this.#fetch;
+  }
+
+  request(url: string, options?: RequestInit): Promise<Response> {
+    const fetch = this.getFetch();
+
+    return this.#enqueue(async () => {
+      const response = await fetch(url, options);
+      // TODO: Check headers
```
SferaDev, line 61: Will follow-up in a new PR

packages/client/src/schema/repository.ts
```diff
@@ -1655,15 +1667,29 @@ export class RestRepository<Record extends XataRecord>
   }
 }
 
-const transformObjectLinks = (object: any) => {
+function cleanSaveRecord<T>(object: any, tableColumns: Schemas.Column[]): T {
```
gimenete, line 1670: I'm not sure about this. The fact that we ignore fields that are not in the database could be a source of confusion.

I think we can maybe expose this function with a different name and allow users to use it explicitely, but not do it internally ourselves.

WDYT?

SferaDev: I would like to avoid adding even more methods, but an option `ignoreExtraProps` might be helpful

gimenete: `ignoreExtraProps` works for me.

PR #726, opened 2022-10-25
[Transactions] Use transactions API internally

SferaDev:
- Use transactions for all `create`, `update` and `delete` operations
  - Transactions is not a real bulk operation
    - Operations that can fail, we query first, remove items that might fail and proceed.
    - To be removed once we have non-transactional APIs in bulk
  - Transactions does not have `columns` for returning projection
    - We are an ORM and users can select column returning
    - We add an extra network call to query the columns projection
- If the provided array has more than the limit of operations (1000), sequentially chunk the requests


packages/client/src/schema/repository.ts
```diff
@@ -1544,6 +1546,35 @@ export class RestRepository<Record extends XataRecord>
...
+        },
+        body: { operations },
+        ...fetchProps
+      });
+
+      return objects;
+    } catch (e) {
+      if (isObject(e) && e.status === 404) {
+        return null;
+      }
```
SferaDev, line 1572: This needs to be updated to the new error the backend returns (in delete might never happen), will add a special type guard to extract transactional errors.

packages/client/src/schema/repository.ts
```diff
@@ -1383,8 +1432,11 @@ export class RestRepository<Record extends XataRecord>
       if (Array.isArray(a)) {
         if (a.length === 0) return [];
 
+        const ids = await this.#insertRecords(a, { ifVersion, createOnly: false });
+
         const columns = isStringArray(b) ? b : (['*'] as K[]);
-        return this.#bulkInsertTableRecords(a, columns);
+        const result = await this.read(ids as string[], columns);
```
SferaDev, line 1438:
```suggestion

        // TODO: Transaction API does not support column projection
        const result = await this.read(ids as string[], columns);
```

tsg (approved):
LGTM

PR #730, opened 2022-10-25
Add check for browser client execution

SferaDev:
Error message:

> You are trying to use Xata from the browser, which is potentially a non-secure environment. If you understand the security concerns, pass `enableBrowser: true` to the client options to remove this error.

### DIFF
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -8,6 +8,7 @@ import { SearchPlugin, SearchPluginResult } from './search';
 import { getAPIKey } from './util/apiKey';
 import { BranchStrategy, BranchStrategyOption, BranchStrategyValue, isBranchStrategyBuilder } from './util/branches';
 import { getCurrentBranchName, getDatabaseURL } from './util/config';
+import { getEnableBrowserVariable } from './util/environment';
 import { getFetchImplementation } from './util/fetch';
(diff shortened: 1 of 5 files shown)


packages/client/src/client.ts
```diff
@@ -76,6 +77,17 @@ export const buildClient = <Plugins extends Record<string, XataPlugin> = {}>(plu
...
 
     #parseOptions(options?: BaseClientOptions): SafeOptions {
+      const enableBrowser = options?.enableBrowser ?? false;
+
+      // If is running from the browser and the user didn't pass `enableBrowser` we throw an error
+      // @ts-ignore Window is not defined in Node
+      const isBrowser = typeof window !== 'undefined' && typeof window.document !== 'undefined';
+      if (isBrowser && !enableBrowser) {
+        throw new Error(
+          'You are trying to use Xata from the browser, which is potentially a non-secure environment. If you understand the security concerns, pass `enableBrowser: true` to the client options to remove this error.'
```
SferaDev, line 87:
```suggestion
          'You are trying to use Xata from the browser, which is potentially a non-secure environment. If you understand the security concerns, such as leaking your credentials, pass `enableBrowser: true` to the client options to remove this error.'
```

gimenete (approved):

PR #735, opened 2022-10-26
Add default value support

SferaDev:
CLI:
- Remove feature flag
- Update to the latest changes

SDK:
- Add E2E test
- Add property to schema type


cli/src/feature-flags.ts
```diff
@@ -1,8 +1,8 @@
-const flags = (process.env.XATA_CLI_FLAGS || '')
+export const flags = (process.env.XATA_CLI_FLAGS || '')
```
gimenete, line 1: Where do you need this variable outside of this file?

gimenete (approved):

PR #756, opened 2022-11-10
Fix update method with id in payload

SferaDev:
https://xata-hq.slack.com/archives/C049MP18MC6/p1667945348721489

### DIFF
--- a/packages/client/src/schema/repository.ts
+++ b/packages/client/src/schema/repository.ts
@@ -1216,7 +1216,7 @@ export class RestRepository<Record extends XataRecord>
   ) {
     const fetchProps = await this.#getFetchProps();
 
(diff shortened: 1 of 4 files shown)


packages/client/src/schema/repository.ts
```diff
@@ -1216,7 +1216,7 @@ export class RestRepository<Record extends XataRecord>
   ) {
     const fetchProps = await this.#getFetchProps();
 
-    const record = transformObjectLinks(object);
+    const { id: _id, ...record } = transformObjectLinks(object);
```
SferaDev, line 1219:
```suggestion
    // Ensure id is not present in the update payload
    const { id: _id, ...record } = transformObjectLinks(object);
```

tsg (approved):

PR #757, opened 2022-11-10
Add fetch pool to avoid rate limits

SferaDev:


packages/client/src/util/fetch.ts
```diff
@@ -11,3 +26,82 @@ export function getFetchImplementation(userFetch?: FetchImpl) {
...
+        }
+      });
+
+    if (this.#queue.length > 0 && this.running + this.started < this.#concurrency) {
+      const next = this.#queue.shift();
+      if (next !== undefined) {
+        this.started++;
+        next();
+      }
+    }
```
gimenete, line 103: I think we can remove the `this.#queue.length > 0` check since you are already checkin if `next !== undefined` later.

Nitpick: lines 98-102 are repeated above. We could extract a function `runNext()` and call it twice. WDYT?

gimenete (approved):
LGTM. Just a couple of non-blocking comments.

packages/client/src/util/fetch.ts
```diff
@@ -11,3 +26,82 @@ export function getFetchImplementation(userFetch?: FetchImpl) {
...
+        this.running--;
+
+        const next = this.#queue.shift();
+        if (next !== undefined) {
+          this.started++;
+          next();
+        }
+      });
+
+    if (this.#queue.length > 0 && this.running + this.started < this.#concurrency) {
```
SferaDev, line 97:
```suggestion
    if (this.running + this.started < this.#concurrency) {
```

PR #759, opened 2022-11-10
[Transactions] Add API-like low level run operation

SferaDev:
Adds a type-safe way to run transactions:

```ts
    const response = await xata.transactions.run([
      { insert: { table: 'teams', record: { name: 'a', index: 0 } } },
      { insert: { table: 'users', record: { id: 'j1', full_name: 'b', index: 1 } } },
      { insert: { table: 'teams', record: { id: 'i1', name: 'b', index: 1 }, createOnly: true } },
      { insert: { table: 'users', record: { id: 'j0', full_name: 'replaced', index: 2 }, ifVersion: 0 } }
    ]);
```


gimenete (commented):
I think we talked about transactions a few times in the past. I was thinking on a different API design. More high level. Something like:

```ts
const transation = xata.transactions.begin();
transaction.users.insert({ name: 'a', index: 0 });
//
await transatcion.commit(); // or run()
```

What do you think?

SferaDev:
Yeah @gimenete, I do plan to add a more high level API in the future. But not until we have 1) queries and 2) references between operations.

So there will be two ways of doing transactions (low and high level), and exactly as you depicted them.

Slack thread: https://xata-hq.slack.com/archives/C025YC7KDPT/p1668102989451149?thread_ts=1668102789.040369&cid=C025YC7KDPT

gimenete:
@SferaDev 🤔 what if we wait to have the high level one? I'm not sure if it's worth the effort having two different APIs. Maybe we can talk about this at some point on zoom.

gimenete (approved):

PR #768, opened 2022-11-15
Add clientName to xata constructor

richardgill:


packages/client/src/client.ts
```diff
@@ -20,12 +20,14 @@ export type BaseClientOptions = {
...
 
-type SafeOptions = AllRequired<Omit<BaseClientOptions, 'branch'>> & {
-  branch: () => Promise<string | undefined>;
-  clientID: string;
-};
+type SafeOptions = AllRequired<Omit<BaseClientOptions, 'branch' | 'clientName'>> &
+  Pick<BaseClientOptions, 'clientName'> & {
+    branch: () => Promise<string | undefined>;
+    clientID: string;
+  };
```
SferaDev, line 30:
```suggestion
type SafeOptions = AllRequired<Omit<BaseClientOptions, 'branch'>> & {
  branch: () => Promise<string | undefined>;
  clientID: string;
};
```

packages/client/src/client.ts
```diff
@@ -92,6 +94,7 @@ export const buildClient = <Plugins extends Record<string, XataPlugin> = {}>(plu
       const apiKey = options?.apiKey || getAPIKey();
       const cache = options?.cache ?? new SimpleCache({ defaultQueryTTL: 0 });
       const trace = options?.trace ?? defaultTrace;
+      const clientName = options?.clientName;
```
SferaDev, line 97:
```suggestion
      const clientName = options?.clientName ?? 'sdk';
```

packages/client/src/api/fetcher.ts
```diff
@@ -139,9 +141,9 @@ export async function fetch<
         body: body ? JSON.stringify(body) : undefined,
         headers: {
           'Content-Type': 'application/json',
-          'User-Agent': `Xata client-ts/${VERSION}`,
           'X-Xata-Client-ID': clientID ?? '',
           'X-Xata-Session-ID': sessionID ?? '',
+          'X-Xata-Agent': `client=TS_SDK; version=${VERSION}; ${clientName ? `service=${clientName};` : ''}`,
```
SferaDev, line 146:
```suggestion
          'X-Xata-Agent': `client=TS_SDK; version=${VERSION}; service=${clientName ?? 'sdk'};`,
```

packages/client/src/api/fetcher.ts
```diff
@@ -139,9 +141,9 @@ export async function fetch<
         body: body ? JSON.stringify(body) : undefined,
         headers: {
           'Content-Type': 'application/json',
-          'User-Agent': `Xata client-ts/${VERSION}`,
           'X-Xata-Client-ID': clientID ?? '',
           'X-Xata-Session-ID': sessionID ?? '',
+          'X-Xata-Agent': `client=TS_SDK; version=${VERSION}; service=${clientName ?? 'sdk'};`,
```
SferaDev, line 146:
```suggestion
          'X-Xata-Agent': `client=TS_SDK; version=${VERSION}; service=${clientName ?? 'unknown'};`,
```

SferaDev:
@richardgill I've triggered the merge on the backend, gladly would include this in this week's release.

Please take a look at my changes, I think they're good enough for now.

packages/client/src/client.ts
```diff
@@ -96,6 +97,7 @@ export const buildClient = <Plugins extends Record<string, XataPlugin> = {}>(plu
       const apiKey = options?.apiKey || getAPIKey();
       const cache = options?.cache ?? new SimpleCache({ defaultQueryTTL: 0 });
       const trace = options?.trace ?? defaultTrace;
+      const clientName = options?.clientName ?? 'sdk';
```
richardgill, line 100: @SferaDev I don't think we want this fallback to `'sdk'`? We could leave it `undefined` or set it to `unknown`. 

But clientName is actually the 'service' e.g. `'frontend' | 'cli' | 'playground'` so saying 'sdk' doesn't make sense here.

SferaDev: I wanted to have it so that in mixpanel we can differentiate from raw API requests done maybe with curl, insomnia, postman and the ones coming from an SDK client. Does that make sense?

packages/codegen/src/codegen.ts
```diff
@@ -306,9 +306,7 @@ export async function generate({
         declarations: [{ name: 'xataWorker', initializer: xataWorkerContent }],
         isExported: true,
         leadingTrivia:
-          language === 'javascript'
-            ? `/** @type { import('@xata.io/client').WorkerRunner<XataClient> } */\n`
-            : undefined,
+          language === 'javascript' ? `\n/** @type { import('@xata.io/client').WorkerRunner<XataClient> } */\n` : '\n',
```
SferaDev, line 309: I took advantage of this open PR for a quick fix in the workers codegen

packages/client/src/client.ts
richardgill: Yes we definitely want that, and we have it via: `X-Xata-Agent=client=TS_SDK;`

https://github.com/xataio/client-ts/pull/768/files#diff-76651c34752b02dfceb46181b8b38ef33ea21122080a391c59b93a19647f178fR139

BTW: This is why I preferred the name `service` -> `clientName`

SferaDev: Okay reverting then, will make it nullable though instead of undefined

SferaDev (approved):

Issue #754, opened 2022-11-08
xata init --db fails to register API key


kostasb:
**Describe the bug**
The "[Project Mode](https://xata.io/docs/cli/getting-started#project-mode)" workflow of the CLI fails to register an API Key despite creating it successfully in the web UI.

To reproduce: 

- don't run xata auth login, otherwise delete ~/.config/xata/credentials, make sure there is no XATA_API_KEY env variable
- In a new project folder (no .env, no .xatarc), do xata init with the db option `xata init --db=...`
- You will be prompted to create a new key in your browser
- Upon creating the key, browser shows success status but the CLI errors and exists
- Note that the fail happens _sporadically_ , I've found that if I am very quick to enter my api key name and hit enter it often works. Exact same workflow, no changes. Browser used in the repro is Firefox 106.0.5 (64-bit) on macos 12.4

![image](https://user-images.githubusercontent.com/15780449/200604126-43c3bb79-08df-4e00-9c77-8dc3e41db270.png)

```
 xata --version
@xata.io/cli/0.10.1 darwin-arm64 node-v18.9.0
```

PR #769, opened 2022-11-15
Wait until a newly created API key is ready to be used

gimenete:
Closes https://github.com/xataio/client-ts/issues/754

We do API requests to the data plane right after creating an API key, and that can be too fast and the API key might not be replicated yet.

With this PR we check up to 10 times and 10 seconds to see if the API is valid after that.

### DIFF
--- a/cli/src/commands/init/index.ts
+++ b/cli/src/commands/init/index.ts
@@ -285,7 +285,9 @@ export default class Init extends BaseCommand {
         return;
       } catch (err) {
         if (err instanceof Error && err.message.includes('Invalid API key')) {
-          this.info('Waiting until the new API key is ready to be used...');
+          if (retries % 2 === 0) {
+            this.info('Waiting until the new API key is ready to be used...');
+          }
           await new Promise((resolve) => setTimeout(resolve, 1000));
         } else {
           throw err;
--- a/cli/src/commands/init/index.ts
+++ b/cli/src/commands/init/index.ts
@@ -233,8 +233,11 @@ export default class Init extends BaseCommand {
 
     if (!apiKey) {
       apiKey = await createAPIKeyThroughWebUI();
+      this.apiKeyLocation = 'new';
       // Any following API call must use this API key
       process.env.XATA_API_KEY = apiKey;
+
+      await this.waitUntilAPIKeyIsValid(workspace, region, database);
     }
     this.info(
       'The fallback branch will be used when you are in a git branch that does not have a corresponding Xata branch (a branch with the same name, or linked explicitly)'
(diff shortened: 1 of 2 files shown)


cli/src/commands/init/index.ts
```diff
@@ -271,6 +274,27 @@ export default class Init extends BaseCommand {
...
+    const xata = await this.getXataClient();
+    const maxRetries = 10;
+    let retries = 0;
+    while (retries++ < maxRetries) {
+      try {
+        await xata.branches.getBranchList({ workspace, region, database });
+        return;
+      } catch (err) {
+        if (err instanceof Error && err.message.includes('Invalid API key')) {
+          this.info('Waiting until the new API key is ready to be used...');
```
exekias, line 288: do we need to tell the user about this? normally this will take 1-2 seconds, maybe only telling them if retries > 2?

cli/src/commands/init/index.ts
```diff
@@ -233,8 +233,11 @@ export default class Init extends BaseCommand {
 
     if (!apiKey) {
       apiKey = await createAPIKeyThroughWebUI();
+      this.apiKeyLocation = 'new';
```
exekias, line 236: is this change intended? I don't see any other reference to `apiKeyLocation`

cli/src/commands/init/index.ts
gimenete: That's a good point. I agree with you but at the same time I don't want the UI to feel stuck if for some reason the replication takes longer.

cli/src/commands/init/index.ts
gimenete: It's intended. It is used to print this error message: https://github.com/xataio/client-ts/blob/697d75938be15e18d05c3a77f87d6ba364c5859b/cli/src/base.ts#L184-L189

For some reason it was missing.

cli/src/commands/init/index.ts
gimenete: What about this? https://github.com/xataio/client-ts/pull/769/commits/f816a77a449e7ccabab89dab11196db2da990b2e

exekias: yeah, I don't know how easy would it be to keep adding dots or similar to make it explicit to the user that the CLI is alive, I'm trying to avoid user questions & uncertainties around this message, but maybe it's not a big deal.

anyway let's not block the PR on this

exekias (approved):

SferaDev (approved):

cli/src/commands/init/index.ts
exekias: oh, that works! 👍 thank you!

gimenete:
@SferaDev that's great thinking. Let's merge this as a quick fix, and I'll think tomorrow about that. Thanks for the review!

PR #784, opened 2022-11-23
Changes to xata branch gathering

SferaDev:
- Send telemetry to mixpanel
- Remove extra round-trip if Xata branch is specified through ENV variable


richardgill (approved):
LGTM but the tests failed.

PR #789, opened 2022-11-25
Add version checks for the CLI

SferaDev:

### DIFF
--- a/cli/src/commands/codegen/index.ts
+++ b/cli/src/commands/codegen/index.ts
@@ -4,6 +4,9 @@ import chalk from 'chalk';
 import { mkdir, writeFile } from 'fs/promises';
 import path, { dirname, extname, relative } from 'path';
 import { BaseCommand, ProjectConfig } from '../../base.js';
+import { readJSON } from '../../utils.js';
+import { VERSION as sdkVersion } from '@xata.io/client';
+import semver from 'semver';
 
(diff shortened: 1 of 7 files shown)


cli/src/commands/codegen/index.ts
```diff
@@ -72,6 +75,24 @@ export default class Codegen extends BaseCommand {
...
+    const installedVersion: string | undefined = packageJSON.dependencies['@xata.io/client'];
+    if (!installedVersion) {
+      return this.error(
+        `Could not find @xata.io/client in your package.json. Please install it to use the codegen command.`
+      );
+    }
+
+    if (!semver.satisfies(installedVersion, `^${sdkVersion}`)) {
+      return this.error(
+        `The installed version of @xata.io/client (${installedVersion}) does not satisfy the required version (${sdkVersion}). Please update your package.json.`
```
SferaDev, line 92:
```suggestion
        `The installed version of the SDK (${installedVersion}) does not satisfy the CLI version (${sdkVersion}). Please update the packages to run the codegen.`
```

SferaDev: I would like to:

- If SDK is outdated, offer to update it with a question
- If CLI is outdated, show command to update (or when we have auto update, offer to update it with a question)

Also:

- Add `--force` to ignore the errors and continue if user knows what they're doing

PR #795, opened 2022-11-28
Add warning for missing region

SferaDev:

### DIFF
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -162,6 +162,16 @@ export abstract class BaseCommand extends Command {
       if (result.success) {
         this.projectConfig = result.data;
         this.projectConfigLocation = search.filepath;
+
+        // Temporal, to be removed in the future when everyone has migrated to the new format
+        if (this.projectConfig.databaseURL) {
+          const { region } = this.parseDatabaseURL(this.projectConfig.databaseURL);
+          if (!region) {
+            throw new Error(
+              `Your databaseURL in the .xatarc file is missing the region. Please update it before continuing.`
+            );
+          }
+        }
       } else {
         this.warn(`The configuration file ${search.filepath} was found, but could not be parsed:`);
         this.printZodError(result.error);


cli/src/base.ts
```diff
@@ -162,6 +162,16 @@ export abstract class BaseCommand extends Command {
       if (result.success) {
         this.projectConfig = result.data;
         this.projectConfigLocation = search.filepath;
+
+        // Temporal, to be removed in the future when everyone has migrated to the new format
+        if (this.projectConfig.databaseURL) {
+          const { region } = this.parseDatabaseURL(this.projectConfig.databaseURL);
+          if (!region) {
+            throw new Error(
+              `Your databaseURL in the .xatarc file is missing the region. Please update it before continuing.`
```
SferaDev, line 171:
```suggestion
              `Your databaseURL in the .xatarc file is missing the region. Please update it and re-run codegen before continuing.`
```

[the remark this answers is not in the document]
tsg: Alternative to providing a docs page, we could ask people to contact support if they get issues. That is probably enough and easier than having a docs page.

SferaDev:
image

cli/src/base.ts
```diff
@@ -162,6 +162,16 @@ export abstract class BaseCommand extends Command {
       if (result.success) {
         this.projectConfig = result.data;
         this.projectConfigLocation = search.filepath;
+
+        // Temporal, to be removed in the future when everyone has migrated to the new format
+        if (this.projectConfig.databaseURL) {
+          const { region } = this.parseDatabaseURL(this.projectConfig.databaseURL);
+          if (region === 'eu-west-1' && !this.projectConfig.databaseURL.includes('eu-west-1')) {
```
gimenete, line 169: I guess we are using `eu-west-1` as a default in `parseDatabaseURL`. What if we remove the default and just check if `region` is not defined?

SferaDev: That would affect also to the SDK, are we good with that?

SferaDev:
@gimenete WDYT now?

gimenete (approved):

cli/src/base.ts
gimenete: Yep. I think it's a good idea to make it mandatory from now on.

PR #804, opened 2022-12-01
Improve workers compilation step

SferaDev:
- Move compile, after watcher is ready
- Add CDN import in case resolve fails (npm install not run recently)
  - This is debatable if it should fail, I'll think about it but for now let's play safe
- Add more logs

### DIFF
--- a/cli/src/workers.ts
+++ b/cli/src/workers.ts
@@ -133,13 +133,13 @@ export async function compileWorkers(file: string): Promise<WorkerScript[]> {
     return [];
   }
 
-  console.log(`[watcher] found ${Object.keys(functions).length} workers in ${file}`);
-
   const versions = await getDependencyVersions();
 
   const compiledWorkers: WorkerScript[] = [];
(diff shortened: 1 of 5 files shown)


cli/src/workers.ts
```diff
@@ -52,14 +55,24 @@ export function buildWatcher<T extends WorkerScript>({
 
   watcher
     .on('add', async (path) => {
-      modules[path] = await compile(path);
+      console.log(`[watcher] add ${path}`);
```
gimenete, line 58: Nitpick, what about having a helper function so we don't repeat `console.log("[watcher] "`?

gimenete (approved):

cli/src/workers.ts
SferaDev: Yeah, I don't like adding those logs there. I am adding a lot of logs everywhere to help debugging with beta users, but this will be removed and refactored in a different way.

PR #835, opened 2022-12-22
Add helpers for serialization 

SferaDev:
- Add `toSerializable`
- Add `toString`


atilafassina (approved):

snide (commented):
I'm not a good one to review this, but just noting that after some help from @SferaDev I got this running in my project and it makes sense. :+1:

PR #860, opened 2023-01-16
[Experimental] Add flag for incremental builds

SferaDev:
We have this feature implemented but without a way to test it with the CLI, add an experimental flag that can be set either via arguments or project config. This will be the default in the future, make it opt-in while we finish testing it internally.


packages/codegen/src/codegen.ts
```diff
@@ -82,7 +101,7 @@ export async function generate({
     sourceFile.addImportDeclaration({
       namedImports: packageImports,
       moduleSpecifier: packageName,
-      leadingTrivia: `// Generated by Xata Codegen ${VERSION}. Please do not edit.\n`
+      leadingTrivia: existingCode ? undefined : `// Generated by Xata Codegen ${VERSION}. Please do not edit.\n`
```
richardgill, line 104: Non-blocking suggestion (or good potential follow up): Should we have comments making it super-clear the parts of the file they should / should not edit?

richardgill (approved):
Left a suggestion - but LGTM

packages/codegen/src/codegen.ts
SferaDev: Yes! That's why I'm removing it if it's being generated over, but I want to draft and see the parts that we would keep or not first before adding potentially other comments.

PR #867, opened 2023-01-23
[Workflow] Add new experimental commands

SferaDev:
- Namespace `api` into a Xata Client in the CLI
- Allow customising `host` in XataClient builder (needed for testing override to dev/staging/localhost)
- Remove branch resolution algorithm and async branches
- Make cache impl optional
- Add boilerplate commands for workflow
- Implement experimental `xata pull`, `xata push`, `xata rebase` and `xata branch`
- Implement `xata diff`
- Remove fallback branch

Postponed:
- Handle current branch


cli/src/base.ts
```diff
@@ -234,7 +231,10 @@ export abstract class BaseCommand extends Command {
       });
     }
 
-    this.#xataClient = new XataApiClient({ apiKey, fetch, host, clientName: 'cli' });
+    const databaseURL = 'https://noop-workspace.cli.xata.sh/db/demo';
+    const branch = 'main';
```
SferaDev, line 235: TODO: Add a proxy for `db` namespace access.

cli/src/base.ts
```diff
@@ -515,14 +515,9 @@ export abstract class BaseCommand extends Command {
...
-  async getCurrentBranchName(databaseURL: string) {
-    const profile = await this.getProfile();
-    return getCurrentBranchName({
-      fetchImpl: fetch,
-      databaseURL,
-      apiKey: profile?.apiKey ?? undefined,
-      clientName: 'cli'
-    });
+  async getCurrentBranchName(_databaseURL: string) {
+    // TODO: FIXME: Get real branch name
```
SferaDev, line 519: TODO: Get from profile

cli/src/commands/push/index.ts
```diff
@@ -64,18 +64,20 @@ export default class Push extends BaseCommand {
...
+      workspace,
+      region,
+      database,
+      branch,
+      migrations: newMigrations.map((migration) => ({
+        id: migration.id,
+        parent: migration.parent,
+        checksum: migration.checksum,
+        operations: migration.operations
+      }))
```
SferaDev, line 79:
```suggestion
      migrations: newMigrations
```

cli/src/commands/push/index.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (flags['dry-run']) {
+      this.log('The following migrations would be pushed:');
+      newMigrations.forEach((migration) => {
+        this.log(`  ${migration.id}`);
+      });
+      return;
+    }
+
+    // we use pushBranchMigrations to push all new migrations at once. The API will verify the migrations
+    // and will fail if a migration in the list can not be applied.
```
SferaDev, line 64:
```suggestion
    // TODO: Check for errors and print them
```

richardgill (approved):
LGTM

exekias:
works like a charm! I tried to use `XATA_BRANCH` env var when doing `git push` but it didn't work. I believe we will need to find ways to avoid user mistakes and make it clear what branch you are talking to

exekias (approved):

SferaDev:
@exekias For these comands we don't use XATA_BRANCH env var right now. I think we need to add some kind of state and probably linked with the current local git branch and make it more obvious yeah. But let's iterate on that before making the feature GA.

Right now you can `xata push branch` as in git commands.

Issue #863, opened 2023-01-19
[xata schema edit] Wrong `defaultValue` on `string` when not defined


fabien0102:
**Describe the bug**

Using `xata edit schema`, when I’m creating a `string` column without providing any `defaultValue`, the value is set as `"undefined"` instead of an empty string (default value in the UI)

kostasb:
This causes creating unique columns with `xata schema edit` to fail (`Error: Internal Error (RequestID: 78447486-c8fa-9b38-8aab-5563ba5d479e)`) because it tries to add defaultValue and uniqueness on the column.

SferaDev:
@fabien0102 and @kostasb Can you give it a try with the next experimental version and see if it's fixed for you?

> npx @xata.io/cli@0.0.0-alpha.vd99266f

kostasb:
> @fabien0102 and @kostasb Can you give it a try with the next experimental version and see if it's fixed for you?
> 
> > npx @xata.io/[cli@0.0.0-alpha.vd99266f](mailto:cli@0.0.0-alpha.vd99266f)

Worked well for me. The CLI requires a Default value when Not null is enabled and I was also able to create a `unique` string column. Thanks!

kostasb:
A followup/implication issue: https://github.com/xataio/client-ts/issues/871

PR #869, opened 2023-01-25
default value for editing undefined string column

richardgill:
Fixes: https://github.com/xataio/client-ts/issues/863

Tested locally for:

Strings with default set / not set
Emails with default set / not set
Links with default not set.


SferaDev (approved):

SferaDev:
We should add a changeset for this change so it goes to the changelog

PR #877, opened 2023-01-30
Move random data to importers

SferaDev:
Expose random data logic for programmatic use. Expected uses:

- Frontend UI
- SDK plugin: `xata.import.random(size, <options>)`


richardgill (approved):
LGTM - thanks for doing this

PR #880, opened 2023-01-31
Update to node 18

SferaDev:


cli/src/commands/dbs/list.test.ts
```diff
@@ -77,25 +77,8 @@ describe('databases list', () => {
...
-          ],
-          [
-            "l",
-            "l",
-            "r",
-          ],
-        ]
-      `);
+      expect(printTable.mock.calls[0][1][0][0]).toBe('test');
+      expect(printTable.mock.calls[0][2]).toEqual(['l', 'l', 'r']);
```
richardgill, line 81: Non blocking:

Before: I understood what was going on.
After: I don't think I'd understand without seeing the old cold first.

cli/src/auth-server.test.ts
```diff
@@ -107,9 +107,7 @@ describe('handler', () => {
     httpHandler(req, res);
 
     expect(res.writeHead).toHaveBeenCalledWith(500);
-    expect(res.end).toHaveBeenCalledWith(
-      'Something went wrong: error:04099079:rsa routines:RSA_padding_check_PKCS1_OAEP_mgf1:oaep decoding error'
-    );
+    expect(res.end).toHaveBeenCalledWith('Something went wrong: error:02000079:rsa routines::oaep decoding error');
```
richardgill, line 110: Non blocking suggestion (I have not run this code).

```suggestion
    expect(res.end).toHaveBeenCalledWith(expect.stringContaining('Something went wrong:'))
    expect(res.end).toHaveBeenCalledWith(expect.stringContaining('decoding error'))
```

richardgill (approved):
Left some comments - approved

cli/src/commands/dbs/list.test.ts
SferaDev: Will revisit these tests in the future to make them more self explanatory

PR #909, opened 2023-02-16
Add column bug fixes

richardgill:
What's new:

- Unique + Not Null errors
- Unique + Default Value errors
- Emails are validated
- Default values for text type allowed
- Setting {unique,notNull,defaultValue} for a type that doesn't support it errors
- notNull and unique are set to `'false' initially.
- No longer able to attempt to edit fields which cannot be edited on an existing column


cli/src/commands/schema/edit.ts
```diff
@@ -734,24 +762,49 @@ function validateOptionalBoolean(value?: string) {
...
+}
+
+function validateNotNull(notNullValue: string | undefined, state: ColumnEditState) {
+  const isNotNull = parseBoolean(notNullValue);
+  if (isNotNull && state.values.type && notNullUnsupportedTypes.includes(state.values.type)) {
+    return `Column type \`${state.values.type}\` does not support \`Not null: true\``;
+  }
+
+  return true;
+}
```
fabien0102, line 786: It’s a lot of logic, hard to say if this is working as expected without tests 🙄 Should we add some? 😁

cli/src/commands/schema/edit.ts
```diff
@@ -33,10 +34,35 @@ type EditableTable = Table & {
...
+  };
+};
+
 const types = ['string', 'int', 'float', 'bool', 'text', 'multiple', 'link', 'email', 'datetime'];
 const typesList = types.join(', ');
 const identifier = /^[a-zA-Z0-9-_~]+$/;
 
+const uniqueUnsupportedTypes: string[] = ['text', 'multiple'];
+const defaultValueUnsupportedTypes: string[] = ['multiple', 'link'];
+const notNullUnsupportedTypes: string[] = defaultValueUnsupportedTypes;
```
SferaDev, line 64: [types] Isn't it the same type as not adding explicitly string[]?

cli/src/commands/schema/edit.ts
```diff
@@ -432,26 +456,69 @@ Beware that this can lead to ${chalk.bold(
...
+          }
         },
         {
           name: 'notNull',
           message: 'Whether the column is not nullable (true/false)',
-          validate(value: string, state: ColumnEditState, item: unknown, index: number) {
-            return validateOptionalBoolean(value);
+          validate(value: string | undefined, state: ColumnEditState, item: unknown, index: number) {
+            if (!isColumnAdded && parseBoolean(value) !== parseBoolean(state.initial.notNull)) {
+              return `Cannot change Not null for existing columns`;
```
SferaDev, line 479: [copy] I noticed in errors you're capitalizing Not null, Default Value, Unique, Description. Probably they should be lowercase inside the error message

SferaDev (approved):
Code LGTM. I tested some of the cases and also work as expected. It's try we don't have tests, at some point we could extract all of this logic into a validation package, I think we have duplicated logic around this in the frontend and SDK.

cli/src/commands/schema/edit.ts
richardgill: I'd attempted to match it to what the user sees on the left of each field (I'd messed up and put a capital V on value though). There are three places they're shown
Screenshot 2023-02-17 at 07 48 52

I'm thinking we should make them all lower camel case in all three places? WDYT?

Screenshot 2023-02-17 at 07 54 31

cli/src/commands/schema/edit.ts
richardgill: Thanks copy paste mistake!

cli/src/commands/schema/edit.ts
SferaDev: Sounds good to me the, if it's a reference to something else maybe we make it look a bit different with chalk instead?

richardgill: We can always follow up with any chalk stuff - I think it's good enough for now

https://github.com/xataio/client-ts/pull/909/commits/d6a37ef8da243c94af633153715c60ddbeba4659

Issue #829, opened 2022-12-20
Request of a dependency is an expression


atilafassina:
**Describe the bug**

When using the Xata client the console is polluted with warnings about the client requiring a dependency via an expression.

```
warn  - ../@xata.io+client@0.21.5_typescript@4.9.4/node_modules/@xata.io/client/dist/index.mjs
Critical dependency: the request of a dependency is an expression
```

**To Reproduce**

Use Xata Client with Next.js for example (or another project running Webpack, I'd assume)

**Expected behavior**

No warnings.

**Software version**

- Next.js 13.0.7
- `@xata.io/client` 0.21.5

SferaDev:
Probably related to having a require in git branch resolution. Will hide it a bit more to see if we can make webpack happy.

SferaDev:
Have tried multiple potential fixes and none worked...

SferaDev:
I think I can remove this warning in the next version 🤞🏻

mattapperson:
So the issue here is `const { execSync } = await import(nodeModule);` as next is transpiring this module and does not support dynamic import statements where the import name is not statically checkable

PR #921, opened 2023-02-23
Simplify branch resolution and plugin builder

SferaDev:
Some refactors and spring cleaning of legacy code.

- In the CLI switch to our plugin mechanism to add future plugins (preparation work for #p-import)
- Make plugin build system synchronous (bug with plugins loaded after invocation)
- Remove `as Config` casting in CLI tests (not needed in oclif v2)
- Make `branch` resolution synchronous (aligned with our new approach for #p-workflow)
- Remove git branch gathering (remove warning given by Next.js)
- Expose customizing `host` in the constructor (needed for local testing)
- Unify `fetchImpl` and `fetch` parameter names (inconsistent names in different parts of the codebase)

Closes https://github.com/xataio/client-ts/issues/829


SferaDev:
Moving to draft again, will do another change here

PR #957, opened 2023-03-22
[Workflow] Make branches verbose

SferaDev:
- Verbose command arguments (no main assumption)
- Confirmation on push

### DIFF
--- a/cli/src/commands/push/index.ts
+++ b/cli/src/commands/push/index.ts
@@ -18,7 +18,7 @@ export default class Push extends BaseCommand<typeof Push> {
   };
 
   static args = {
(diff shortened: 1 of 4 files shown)


cli/src/commands/push/index.ts
```diff
@@ -52,14 +53,24 @@ export default class Push extends BaseCommand<typeof Push> {
       return;
     }
 
+    newMigrations.forEach((migration) => {
+      this.log(`  ${migration.id}`);
+    });
+
     if (flags['dry-run']) {
       this.log('The following migrations would be pushed:');
```
SferaDev, line 61:
```suggestion
      this.log(`A total of ${newMigrations.length} migrations would be pushed to ${branch}.`);
```

PR #980, opened 2023-04-20
[Workflow] Remove `git` integration in CLI (1)

SferaDev:
Part 1 of Workflow merge:

- Remove git functionality from create command, default is now the same as --no-git previous behavior


richardgill (approved):
LGTM

PR #987, opened 2023-04-20
[Workflow] Fix `diff` command on fish shell (8)

SferaDev:
Part 8 of workflow merge:

- Fix `diff` command in fish shell
- Do not show `Create database` with `diff` command as it doesn't make sense


richardgill (approved):
LGTM

PR #988, opened 2023-04-20
[Workflow] Remove plural commands

SferaDev:
@richardgill Raised that we have low usage on `workspaces` and `branches` commands and instead of deprecating the commands we could just removing.

Since our decision yesterday was to `deprecate` we have merged the PRs with that decision, but opening this PR to review it.

Pros of removing:

- Less maintenance (even if we will eventually delete them)

Cons of removing:

- Users that were used to the previous command won't be able to use them

I personally think, this sets a bit of precendent regarding CLI breaking changes on documented commands, but since there's low usage and we're still in v0.x I would be fine with whatever choice we make.


SferaDev:
Decision to wait for 2 months and then remove the commands

SferaDev:
2 months reached. This will be included in the next version.

richardgill (approved):
LGTM

PR #993, opened 2023-04-27
Add `xata.*` columns to `SelectableColumn`

andrew-farries:

### DIFF
--- a/.changeset/orange-hornets-appear.md
+++ b/.changeset/orange-hornets-appear.md
@@ -1,0 +1,6 @@
+---
+"@xata.io/cli": patch
+"@xata.io/client": patch
+---
+
+Add `xata.*` columns to `SelectableColumn`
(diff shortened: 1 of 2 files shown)


SferaDev:
Not sure about adding it to selectable columns as they are completely virtual, probably they should be added to the `metadata` similar to how `version` or the other properties coming from search can be retrieved

andrew-farries:
> Not sure about adding it to selectable columns as they are completely virtual, probably they should be added to the `metadata` similar to how `version` or the other properties coming from search can be retrieved

Thanks. I haven't worked with the Typescript SDK before, so this draft PR is just me playing around with how we could add this 😄

SferaDev:
If you want we can pair and implement it together. There're 2 things you need to make sure here:

1. Types
2. Actual response matches the types

There's a way to increase the types coming from the metadata of a XataRecord via it's generic type.

andrew-farries:
Closing in favour of https://github.com/xataio/client-ts/pull/1009

PR #1008, opened 2023-05-17
Add `failIfMissing` delete option

SferaDev:
- Re-enable tests for `transactions` (we disabled them a while back because the columns breaking change, updated to add them to the snapshots)
- Add `failIfMissing` option
- Add test for `failIfMissing`


test/integration/transactions.test.ts
```diff
@@ -285,4 +285,19 @@ describe('delete transactions', () => {
 
     await xata.db.teams.delete('i1');
   });
+
+  test('delete with failIfMissing', async () => {
+    const records = await xata.transactions.run([{ delete: { table: 'teams', id: 'ab' } }]);
```
SferaDev, line 290:
```suggestion
    const records = await xata.transactions.run([{ delete: { table: 'teams', id: 'ab', failIfMissing: false } }]);
```

richardgill (approved):
LGTM

PR #1012, opened 2023-05-19
Add `factor` to date booster

SferaDev:


tsg (approved):

tsg:
@SferaDev after this is merged, could you also cut a release so that we can pass the fix to Raae?

PR #1018, opened 2023-05-24
bugfix: CLI localhost:port/new redirects

richardgill:
Screenshot 2023-05-24 at 14 48 18

Clicking this link sends you to a url like this:

https://app.xata.io/new-api-key?pub=key&name=Xata+CLI&redirect=http%3A%2F%2Flocalhost%3A80

But the redirect url was always http://localhost:80 which is wrong and fails the flow completely.

(I was trying to paste it into an incognito tab for reasons)

Test with: `npx @xata.io/cli@0.0.0-alpha.vc5b0c25 auth login` and click the link

### DIFF
--- a/cli/src/auth-server.ts
+++ b/cli/src/auth-server.ts
@@ -20,7 +20,7 @@ export function handler(publicKey: string, privateKey: string, passphrase: strin
 
       const parsedURL = url.parse(req.url ?? '', true);
       if (parsedURL.pathname === '/new') {
-        const port = +(parsedURL.port || 80);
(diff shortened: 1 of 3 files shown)


cli/src/auth-server.ts
```diff
@@ -20,7 +20,7 @@ export function handler(publicKey: string, privateKey: string, passphrase: strin
 
       const parsedURL = url.parse(req.url ?? '', true);
```
richardgill, line 21: `req.url` looks like `"/new"` which has no port!

SferaDev (approved):

PR #1019, opened 2023-05-24
`xata init` improvements and fixes

richardgill:

### DIFF
--- a/cli/src/commands/init/index.ts
+++ b/cli/src/commands/init/index.ts
@@ -4,6 +4,7 @@ import { ModuleType } from '@xata.io/codegen';
 import chalk from 'chalk';
 import dotenv from 'dotenv';
 import { access, readFile, writeFile } from 'fs/promises';
+import compact from 'lodash.compact';
 import path, { extname } from 'path';
 import which from 'which';
 import { createAPIKeyThroughWebUI } from '../../auth-server.js';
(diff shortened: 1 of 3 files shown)


snide (approved):
Tested this ran locally as before and included the text changes. LGTM.

cli/src/commands/init/index.ts
```diff
@@ -209,9 +215,10 @@ export default class Init extends BaseCommand<typeof Init> {
       return { name: 'yarn', args: ['add'] };
     } else if (await this.access('pnpm-lock.yaml')) {
       return { name: 'pnpm', args: ['add'] };
-    } else {
-      return null;
+    } else if (await this.access('package.json')) {
+      return { name: 'npm', args: ['install', '--save'] };
```
SferaDev, line 219: I'm not sure of this change, if you don't have a package manager configured I wouldn't force it to npm.

Can we add a new question at this stage? This should barely happen and I would like to avoid that we do unwanted changes to a users codebase.

richardgill: So it happened to me yesterday, like this: `pnpm init && xata init`

`pnpm init` doesn't create a lock file, only a `package.json`.

So I think for people who are just trying us out this might be more common than we think.

richardgill: @SferaDev I've backed out this change in this PR, will follow up with a a PR to add an extra question for this code path 👍

SferaDev (approved):

PR #1021, opened 2023-05-25
`xata init` improvements

richardgill:

### DIFF
--- a/cli/src/commands/init/index.ts
+++ b/cli/src/commands/init/index.ts
@@ -11,6 +11,7 @@ import { createAPIKeyThroughWebUI } from '../../auth-server.js';
 import { BaseCommand, ENV_FILES } from '../../base.js';
 import { isIgnored } from '../../git.js';
 import { xataDatabaseSchema } from '../../schema.js';
+import { delay } from '../../utils/delay.js';
 import Browse from '../browse/index.js';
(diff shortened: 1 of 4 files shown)


cli/src/commands/init/index.ts
```diff
@@ -188,33 +220,41 @@ export default class Init extends BaseCommand<typeof Init> {
 
   async getPackageManager() {
     const packageManager = await this.guessPackageManager();
```
SferaDev, line 222: Should we use `flags.packageManager` here as a default?

richardgill: fixed with https://github.com/xataio/client-ts/pull/1021/commits/2f010952c16852653f6fd00f27c7928587d5e826

SferaDev (approved):
LGTM. Added a proposal to build a helper for oclif enum flags in https://github.com/xataio/client-ts/pull/1023. I think this could come handy here and in other commands aswell.

cli/src/commands/init/index.ts
```diff
@@ -50,6 +78,10 @@ export default class Init extends BaseCommand<typeof Init> {
     }),
     schema: Flags.string({
       description: 'Initializes a new database or updates an existing one with the given schema'
+    }),
+    packageManager: Flags.string({
```
SferaDev, line 82: Almost forgot! This should be `package-manager` if we follow CLI flag naming conventions

PR #1023, opened 2023-05-26
Add helper for enum flags in CLI

SferaDev:


cli/src/utils/oclif.ts
```diff
@@ -0,0 +1,10 @@
+import { Flags } from '@oclif/core';
+import { FlagProps } from '@oclif/core/lib/interfaces/parser';
+
+type EnumFlagOptions<Enum> = Omit<FlagProps, 'name'> & {
+  options: (Enum | string)[];
```
SferaDev, line 5: The `Enum | string` is a trick that accomplished a two-fold inference solution.

You can do:

```ts
packageManager: enumFlag<PackageManageKey>({
  description: 'The package manager to use to install the @xata.io/client package',
  options: Object.keys(packageManagers)
})
```

or:

```ts
packageManager: enumFlag({
  description: 'The package manager to use to install the @xata.io/client package',
  options: ['npm', 'pnpm', 'yarn'] as const
})
```

And in both cases the inferred type is: 'npm' | 'pnpm' | 'yarn'.

PR #1024, opened 2023-05-26
Don't include xata object in not expanded links

SferaDev:
Fixes https://github.com/xataio/client-ts/issues/1047


richardgill (approved):
LGTM

PR #1029, opened 2023-05-31
Xata init quality improvements

richardgill:
@SferaDev I've implemented your improvements from #1021 👍


cli/src/utils/mockFs.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  return matchingDirs;
+};
+
+const NODE_MODULES_PATH = '../../../node_modules/.pnpm';
+const ABSOLUTE_MODULE_PATH = path.resolve(__dirname, NODE_MODULES_PATH);
+const COSMICONFIG_PATH = findDirectories(ABSOLUTE_MODULE_PATH, /cosmiconfig@.*/)[0];
+const OCLIF_PATH = findDirectories(ABSOLUTE_MODULE_PATH, /@oclif\+core@.+_@types\+node@.+_typescript@.+/)[0];
+
+export const mockFileSystem = (files: Record<string, string>) => {
```
richardgill, line 37: I tried very hard to use memfs but it doesn't work for node_modules which use `fs` internally. This solution seems to work well and runs fast. 

Here is how the code looked just as I decided to switch: https://github.com/xataio/client-ts/compare/main...xata-init-quality-improvements-2-3backup

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "vitest",
+    "test": "vitest --no-threads",
```
richardgill, line 7: change do `process.chdir(` without `--no-threads`  https://github.com/vitest-dev/vitest/issues/566

cli/src/base.test.ts
```diff
@@ -12,9 +12,7 @@ class FakeCommand extends BaseCommand<typeof FakeCommand> {
   }
 }
 
-beforeEach(() => {
-  clearEnvVariables();
-});
+clearEnvVariables();
```
richardgill, line 15: @SferaDev this was what was breaking the tests

richardgill:
@SferaDev I've added a couple more commits from your last review which don't really touch the implementation.

1 test is still failing, but I reran a successful test from earlier and it failed with the same error - so I don't think it's me: 

https://github.com/xataio/client-ts/actions/runs/5131940326/jobs/9266326209

SferaDev (approved):

PR #1030, opened 2023-05-31
Make `xata` property enumerable

### LINKED ISSUE #1027: xata property not available in console output
**Describe the bug**
The xata property (built-in property that provides version, createdAt, updatedAt) is not available in the console output.

To reproduce, log output from any get method (getFirst, getMany etc) and notice that the xata property is not part of it:

```
const record = await xata.db.table.getFirst();
console.log(record);
```

Reported with version: 0.24.0

SferaDev:
- Make sure it's enumerable (so it appears on console.log and object cloning)
- Freeze the object
- Add more tests

Fixes #1027

### DIFF
--- a/.changeset/beige-donkeys-own.md
+++ b/.changeset/beige-donkeys-own.md
@@ -1,0 +1,5 @@
+---
+"@xata.io/client": patch
+---
+
+Make `xata` property enumerable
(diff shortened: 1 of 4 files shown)


richardgill (approved):
LGTM

PR #1035, opened 2023-06-02
Simplify test setup

SferaDev:
Use newer APIs and reduce number of calls

### DIFF
--- a/packages/codegen/example/schema.json
+++ b/packages/codegen/example/schema.json
@@ -1,5 +1,4 @@
 {
-  "formatVersion": "1.0",
   "tables": [
     {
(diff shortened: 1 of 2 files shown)


packages/codegen/example/schema.json
```diff
@@ -71,6 +70,14 @@
...
           "type": "string"
         },
+        {
+          "name": "photo",
+          "type": "file"
+        },
+        {
+          "name": "attachments",
+          "type": "file[]"
+        },
```
SferaDev, line 80:
```suggestion
```

richardgill (approved):
LGTM

PR #1060, opened 2023-07-04
Unify place to add new column types

SferaDev:
Right now we need to add new column types in multiple places, add a common way to add them and import


packages/codegen/package.json
```diff
@@ -31,6 +31,7 @@
   },
   "homepage": "https://github.com/xataio/client-ts/blob/main/codegen/README.md",
   "dependencies": {
+    "@xata.io/client": "workspace:*",
```
richardgill, line 34: how come this changed?

richardgill (approved):
LGTM

Left a non-blocking question

packages/codegen/package.json
SferaDev: We didn't use anything from the client before and we are now using it.

PR #1061, opened 2023-07-04
Update to pnpm v8

SferaDev:


cli/package.json
```diff
@@ -55,6 +55,7 @@
     "open": "^9.1.0",
     "prompts": "^2.4.2",
     "relaxed-json": "^1.0.3",
+    "rollup": "^3.26.0",
```
richardgill, line 58: how come this changed?

richardgill (approved):
@SferaDev explained that it's expected that bundle size will fail here.

LGTM

cli/package.json
SferaDev: pnpm v8 changes the way peer deps are handled. In v7 we didn't add it as a dependency ourselves and it was picking one because of a peer, but with v8 failed because we forgot to add it to our package.json

PR #1063, opened 2023-07-04
Match column selection with paths

SferaDev:
Column selection only matched exact values or links, for objects or files it didn't expand column selection.


richardgill (approved):
lGTM

PR #1067, opened 2023-07-04
Fix partial filters returning 400

### LINKED ISSUE #1058: Multi-clause filter errors with void clause parameter
**Describe the bug**

Tested with 0.24.3

When applying multiple clauses in one filter, calling an optional non-existent field creates a void clause that yields `"filter invalid: filter [$all]: filter empty"`, even if there are other valid clauses in the array:

```ts
const session = { user: { email: "someone@example.com" } };

const records = await xata.db.table1.filter({
    s1: session?.user?.notemail,
    s2: "a",
}).getMany();
```

REST:
```
{"filter":{"$all":[{},{"s2":"a"}]}
```

While, chaining filters ignores the void filter which avoids the empty filter error:

```ts
const records = await xata.db.table1
  .filter("s1", session?.user?.notemail)
  .filter("s2", "a")
  .getMany()
```

REST:
```
{"filter":{"$all":[{"s2":"a"}]}
```

It would be nice to align the behavior of multi-clause syntax with chained filters.

SferaDev:
Fix #1058

### DIFF
--- a/packages/client/src/schema/filters.ts
+++ b/packages/client/src/schema/filters.ts
@@ -1,5 +1,5 @@
 import { FilterExpression, FilterPredicate } from '../api/schemas';
-import { isDefined } from '../util/lang';
+import { isDefined, isObject } from '../util/lang';
 import { SingleOrArray } from '../util/types';
 import { XataRecordMetadata } from './record';
 import { ColumnsByValue, ValueAtColumn } from './selection';
@@ -129,6 +129,14 @@ export function cleanFilter(filter?: FilterExpression | FilterPredicate): Filter
 
         // Remove empty arrays
         if (clean.length === 0) return acc;
+
+        return [...acc, [key, clean]];
+      }
+
+      if (isObject(value)) {
+        // Remove empty objects
+        const clean = cleanFilter(value);
+        if (!isDefined(clean)) return acc;
 
         return [...acc, [key, clean]];
       }
--- a/packages/client/src/schema/filters.ts
+++ b/packages/client/src/schema/filters.ts
@@ -115,8 +115,9 @@ export type Filter<T> = T extends Record<string, any>
     : BaseApiFilter<T> | NestedApiFilter<T>
   : PropertyFilter<T>;
 
-export function cleanFilter(filter?: FilterExpression | FilterPredicate): FilterExpression | undefined {
+export function cleanFilter(filter?: FilterExpression | FilterPredicate): any {
   if (!isDefined(filter)) return undefined;
+  if (!isObject(filter)) return filter;
 
   const values = Object.fromEntries(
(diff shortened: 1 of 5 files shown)


packages/client/src/schema/filters.ts
```diff
@@ -114,12 +115,27 @@ export type Filter<T> = T extends Record<string, any>
...
-    .filter((value) => (Array.isArray(value) ? value.length > 0 : true));
+  const values = Object.fromEntries(
+    Object.entries(filter).reduce((acc, [key, value]) => {
+      // Remove null and undefined values
+      if (!isDefined(value)) return acc;
 
-  return values.length > 0 ? filter : undefined;
+      if (Array.isArray(value)) {
+        // Remove empty objects from arrays
+        const clean = value.map((item) => cleanFilter(item)).filter((item) => isDefined(item));
```
SferaDev, line 128: This is the real change, for arrays we now iterate and clean the values from undefined and nulls. The rest of diff is changing from `filter` to `reduce` to do the same checks.

packages/client/src/schema/filters.ts
```diff
@@ -114,12 +115,27 @@ export type Filter<T> = T extends Record<string, any>
...
-  const values = Object.values(filter)
-    .filter(Boolean)
-    .filter((value) => (Array.isArray(value) ? value.length > 0 : true));
+  const values = Object.fromEntries(
+    Object.entries(filter).reduce((acc, [key, value]) => {
+      // Remove null and undefined values
+      if (!isDefined(value)) return acc;
 
-  return values.length > 0 ? filter : undefined;
+      if (Array.isArray(value)) {
```
richardgill, line 126: Do we have a test for this `if` branch of code? 

I can't see how the test you added covers it, but perhaps there is a prexisting one?

I think we should have a test the `if (clean.length === 0) return acc;` case and the `return [...acc, [key, clean]];` case?

packages/client/src/schema/filters.ts
```diff
@@ -114,12 +115,27 @@ export type Filter<T> = T extends Record<string, any>
...
+        if (clean.length === 0) return acc;
+
+        return [...acc, [key, clean]];
+      }
+
+      return [...acc, [key, value]];
+    }, [] as [string, any][])
+  );
+
+  return Object.keys(values).length > 0 ? values : undefined;
```
richardgill, line 140: Do we have a test for `Object.keys(values).length > 0 === false` case?

richardgill (commented):
Change looks good, I've left some comments about testing.

SferaDev:
@richardgill Added unit tests instead of integration ones

richardgill (approved):
LGTM - thanks for the extra tests 👍

PR #1076, opened 2023-07-12
Importer CLI improvements

SferaDev:


kostasb:
Tried it out, some feedback:

with the existing table schema having brokerage as link, the CLI suggests to remove the link column (brokerage)

```
id,licenseNumber,licenseHolderName,licenseType,stateLicenseType,brokerage,licenseExpirationDate,address1,address2,city,county,zip,state
1,license1,name1,type1,state1,broker1,2023-07-10T16:04:33.932152Z,addr1,addr2,ny,us,16232,california
```

```
xataCli import csv agentities.csv --table=agent
✖ WARNING: The following columns will be removed and you will lose data. brokerage. 
Do you want to continue? …
```

SferaDev:
Yeah, without the flag it will try to remove the `link` column and create an `int` one, this is the new default behaviour that @richardgill implemented with the new importer. There could be a discussion about if that should be the default or not, no strong preferences in this case, up to @richardgill

kostasb:
We can provide a hint to use the flag `keep-existing-columns` in this warning message then?

PR #1078, opened 2023-07-12
Add links to the CLI

richardgill:


cli/src/commands/import/csv.ts
```diff
@@ -267,6 +272,9 @@ const flagsToColumns = (flags: {
   }
   return columns.map((name, i) => {
     const type = importColumnTypes.parse(types[i]);
+    if (type === 'link') {
+      return { name, type, link: { table: name } };
```
richardgill, line 276: if the table name doesn't match the schema APIs error ~sensibly

SferaDev (approved):

PR #1087, opened 2023-07-21
Import: row indexes, uncoerced-data & errors

richardgill:


richardgill:
Renamed to `original`, I think that's a better name.

SferaDev (approved):

PR #1091, opened 2023-07-24
Cli import multiple

richardgill:
Support importing multiple.

Tested with `--types` flag and inferring a multiple.

https://github.com/xataio/file-import-samples updated to include multiples

### DIFF
--- a/packages/importer/src/columns.ts
+++ b/packages/importer/src/columns.ts
@@ -91,38 +91,41 @@ export const guessColumnTypes = <T>(
   return 'string';
 };
 
+export type CoercedValue = { value: string | number | boolean | Date | null; isError: boolean };
+
 export const coerceValue = (
   value: unknown,
   type: Schemas.Column['type'],
   options: ColumnOptions = {}
-): string | number | boolean | Date | null => {
+): CoercedValue => {
   const { isNull = defaultIsNull, toBoolean = defaultToBoolean } = options;
 
   if (isNull(value)) {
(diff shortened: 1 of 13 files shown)


packages/importer/src/columns.ts
```diff
@@ -84,14 +114,17 @@ export const guessColumnTypes = <T>(
   if (columnValues.every(isEmail)) {
     return 'email';
   }
+  if (columnValues.some(isMultiple)) {
```
richardgill, line 117: A good option might be to just never guess `multiple`

SferaDev: I'm okay not guessing it, if uploading with an existing schema works or passing the type explicitly works.

SferaDev:
@kostasb Can you give it a try?

SferaDev (approved):

kostasb:
Works well when explicitly specifying the type as multiple!

Not sure if we were planning to infer multiples or just create string columns if not specifying? It doesn't seem to infer multiple.

```
id,mymulti,mystring
1,"one,two",three

xataCli import csv my.csv --table=mytable1 --branch=main --columns=id,mymulti,mystring --types=string,multiple,string
// works!

xataCli import csv my.csv --table=mytable2 --branch=main
// mymulti is created as a string column
```

SferaDev:
> It doesn't seem to infer multiple.

We discussed this with Richard and no infer for now as it will detect some false positives

Issue #992, opened 2023-04-26
xata init --force should clear migration files


kostasb:
**Describe the bug**
When running `xata init --force` the command should take care of cleaning up, or archiving in some way so they don't get in the way, any existing migration files.

Use case: changing the database associated with a project by initializing with a new db. Currently, the user needs to zero off the .xata/migrations directory otherwise the projects ends up in a conflicting state.

**To Reproduce**

The following script is with `@xata.io/cli/0.11.19`

```
//in a project with a db configured already, having migration files
my-xata-app % ls .xata/migrations 
mig_ch4j00rfqdrklh9k3t1g_a5ebae2a.json	mig_ch4j00rfqdrklh9k3t2g_87ff5a35.json

my-xata-app % xata init --force
...configure a new db

my-xata-app % xata pull main
There are new migrations on both the local and remote branches. Please run `xata rebase` to resolve the conflicts.

my-xata-app % xata rebase main
    Error: ENOENT: no such file or directory, open '/Users/kostas/Documents/pworkflow/my-xata-app/.xata/migrations/mig_cgvs0dr8ecc81uop9jb0_65112536.json'
    Code: ENOENT
```

SferaDev:
@kostasb https://github.com/xataio/client-ts/pull/1092 Can you check different scenarios with this change? I will also test it out when I get a chance.

PR #1092, opened 2023-07-25
Run pull on init

SferaDev:
Fixes https://github.com/xataio/client-ts/issues/992


kostasb:
While init-ing on a fresh directory, there's an error due to the migrations dir not being there

> no such file or directory, stat '/Users/kostas/testdir/.xata/migrations'


```
xataCli init
Need to install the following packages:
  @xata.io/cli@0.0.0-alpha.v38ffdc3
Ok to proceed? (y) y
🦋 Initializing project... We will ask you some questions.

✔ Select a workspace › ws
✔ Select a database or create a new one › test
✔ Create .gitignore and ignore .env? … yes
✔ Generate code and types from your Xata database › TypeScript
✔ Choose the output path for the generated code … src/xata.ts

Setting up Xata...

Created Xata config: .xatarc

Creating .env file
  set XATA_API_KEY=xau_*********************************
  set XATA_BRANCH=main

Added .env file to .gitignore

    Error: ENOENT: no such file or directory, stat '/Users/kostas/testdir/.xata/migrations'
    Code: ENOENT

```

kostasb:
Didn't step on any other issues with this approach, pull on init works well (as long as the .xata/migrations directory exists)

SferaDev:
Okay will fix the problem with directory

SferaDev:
@kostasb Can you try again when the new build finishes?

kostasb:
> @kostasb Can you try again when the new build finishes?

Worked well this time, thanks!

cli/src/commands/init/index.ts
```diff
@@ -158,6 +159,8 @@ export default class Init extends BaseCommand<typeof Init> {
       await this.deploySchema(workspace, region, database, branch, schema);
     }
 
+    // Run pull to retrieve remote migrations, remove any local migrations, and generate code
```
richardgill, line 162: This comment made me think pull generates code? 

If it does, should we remove the codegen line?

richardgill (approved):
Left a comment.

LGTM

cli/src/commands/init/index.ts
SferaDev: It does but the config is not saved yet, so we need to call them both.

PR #1094, opened 2023-07-25
`xata.io/importer`: batch import API change

richardgill:


packages/importer/src/types.ts
```diff
@@ -178,7 +178,7 @@ export type CsvResults = {
 export type ImportBatchOptions = {
   columns: Column[];
   table: string;
-  batch: ParseResults;
+  batchRows: unknown[];
```
richardgill, line 181: This is really the change

SferaDev (approved):

PR #1098, opened 2023-07-28
Importer: Allow commas in numbers

richardgill:


SferaDev (commented):
As discussed over slack, the `replaceAll` solution doesn't work in European countries. We should either add some more complex algorithm to detect them (and know that there're some false positives around it) or at least make the behaviour configurable and set defaults based on user locale.

PR #1101, opened 2023-07-28
Add cancellable timeout

SferaDev:
Found by a user in discord, CLI was waiting for timeout of last fetch call to settle.


snide (approved):
Verified that this fixes the lag I could replicate with the newest client.

PR #1108, opened 2023-07-31
Importer trim whitespace

richardgill:
Handle whitespace better for int,float,datetime,email column types.


SferaDev (approved):
Please add a changeset and we're good to go

PR #1114, opened 2023-08-03
Fix warning with default branch

SferaDev:
We were defaulting to main on env branch and that was showing an invalid warning.

Checked the usage of `getBranch` function and the behavior change is safe.


richardgill (approved):
LGTM

PR #1122, opened 2023-08-07
Add support for OAuth login

SferaDev:


cli/src/base.ts
```diff
@@ -188,33 +191,57 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
...
+      token: token ?? credential?.apiKey ?? ''
+    };
   }
 
   async getXataClient({ profile }: { profile?: Profile } = {}) {
     if (this.#xataClient) return this.#xataClient;
 
-    const { apiKey, host } = profile ?? (await this.getProfile());
-
-    if (!apiKey) {
```
eemmiillyy, line 210: Should we still check whether the token is an empty string and log that error?

cli/src/base.ts
```diff
@@ -538,29 +565,11 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
...
-        message: 'Existing API key:'
-      });
-      if (!key) this.exit(2);
-      return key;
-    }
+  async obtainKey(web?: string) {
+    const currentProfile = await this.getProfile();
+    const domain = web || currentProfile?.web || 'https://app.xata.io';
+
+    return await loginWithWebUI(domain);
```
eemmiillyy, line 572: We no longer give them the option to create a new API key during login, it's done automatically on login now?

cli/src/commands/auth/refresh.ts
```diff
@@ -0,0 +1,45 @@
+import { refreshAccessToken } from '../../auth-server.js';
+import { BaseCommand } from '../../base.js';
+import { readCredentialsDictionary, saveCredentials } from '../../credentials.js';
+
+export default class Refresh extends BaseCommand<typeof Refresh> {
```
eemmiillyy, line 5: When is this called?

cli/src/auth-server.ts
```diff
@@ -37,38 +58,45 @@ export function handler(publicKey: string, privateKey: string, passphrase: strin
...
     .replace('-----END PUBLIC KEY-----', '');
-  const name = 'Xata CLI';
-  const redirect = `http://localhost:${port}`;
-  const url = new URL('https://app.xata.io/new-api-key');
-  url.searchParams.append('pub', pub);
-  url.searchParams.append('name', name);
-  url.searchParams.append('redirect', redirect);
+
+  const url = new URL(`${domain}/integrations/oauth/authorize`);
+  url.searchParams.set('client_id', 'kav7lscnhp0k5a608rg58fjf6g');
```
SferaDev, line 92:
```suggestion
  url.searchParams.set('client_id', 'b7msdmpun91q33vpihpk3vs39g');
```

PR #1140, opened 2023-08-30
[Attachments] Add metadata url for transformations

SferaDev:
When resizing or applying other transformations, we alter the metadata of the image. Cloudflare offers a way to get the metadata both of the original and transformed image. Expose that URL to the user in case they want to fetch it.


packages/client/src/schema/files.ts
```diff
@@ -165,7 +165,8 @@ export class XataFile {
   public transform(...options: ImageTransformations[]) {
     return {
       url: transformImage(this.url, ...options),
-      signedUrl: transformImage(this.signedUrl, ...options)
+      signedUrl: transformImage(this.signedUrl, ...options),
+      metadataUrl: transformImage(this.url, ...options, { format: 'json' })
```
SferaDev, line 169: Not sure if it's worth adding a `metadataSignedUrl`, normally `signedUrl` is used for displaying it in the DOM not for programmatic access.

richardgill (approved):

PR #1145, opened 2023-08-31
[SQL] Add driver for drizzle

SferaDev:


packages/plugin-client-drizzle/README.md
```diff
@@ -0,0 +1,59 @@
...
+const xata = getXataClient();
+const db = drizzle(xata);
+
+const drivers = pgTable('drivers', {
+  id: text('id').primaryKey(),
+  surname: text('surname'),
+  forename: text('forename')
+});
+
+const result = await db.select().from(drivers).execute();
```
tsg, line 40: nit: Could we show a where clause here?

packages/plugin-client-drizzle/README.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+## Model generation
+
+We offer an experimental model generation helper that will generate the models for you from your `tables` array in your `xata.ts` file.
```
tsg, line 45: Ooooh, cool

packages/plugin-client-drizzle/src/generate.ts
```diff
@@ -0,0 +1,65 @@
...
+    case 'int':
+      return integer(column.name);
+    case 'float':
+      return decimal(column.name);
+    case 'datetime':
+      return date(column.name);
+    case 'multiple':
+    case 'object':
+    case 'vector':
+    case 'file[]':
```
tsg, line 23: Ah, now I understand you slack question.`vector` is "array of `real`" but not sure have that is represented. `file[]` is array of `jsonb`.

packages/plugin-client-drizzle/src/generate.ts
```diff
@@ -0,0 +1,65 @@
...
+    case 'file':
+    case 'json':
```
tsg, line 25: file and json can be mapped to `jsonb`.

tsg (approved):
LGTM. I expect the model generation will break on objects and some of the more complex types, so I think in the docs we should recommend manually creating the schema for now.

packages/plugin-client-drizzle/README.md
tsg: `  .where(eq('nationality', 'spanish'))` I think should work, with `eq` imported from `drizzle-orm`.

packages/plugin-client-drizzle/src/generate.ts
SferaDev: I'll leave them as TODO, it's experimental and it's not clear how they'll map to the ORM

packages/plugin-client-drizzle/README.md
SferaDev: Updated with a more complex example

PR #1156, opened 2023-09-05
[Files] Add download transformation

SferaDev:


packages/client/src/files/transformations.ts
```diff
@@ -49,6 +49,11 @@ export interface ImageTransformations {
    * ignored.
    */
   contrast?: number;
+  /**
+   * Download file. Forces browser to download the image.
+   * Value is used for the download file name.
```
SferaDev, line 54:
```suggestion
   * Value is used for the download file name. Extension is optional.
```

SferaDev:
@sorintm Can you review it? 😄

sorintm (approved):
LGTM!

sorintm:
(Not directly related to this PR) @SferaDev do you think we should have some validation or sanitization of the values we accept for file name in this transformation?
I couldn't think of attack vectors as we just put the value in a Response header. 
I think the problem would be if the client file system does not accept the name as valid, but that should be an obvious user error.

SferaDev:
I don't think it's necessary, but if you think we should I'm okay adding it

PR #1161, opened 2023-09-07
Migrate to global fetch in node

SferaDev:
Since we now target node 18+, remove `node-fetch` dependency from tests. We still keep it in the CLI for now although we might want to reconsider that soon.


packages/client/src/util/fetch.ts
```diff
@@ -66,7 +72,10 @@ export class ApiRequestPool {
     const runRequest = async (stalled = false): Promise<Response> => {
       // Some fetch implementations don't timeout and network changes hang the connection
       const { promise, cancel } = timeoutWithCancel(REQUEST_TIMEOUT);
-      const response = await Promise.race([fetchImpl(url, options), promise.then(() => null)]).finally(cancel);
+      const response = await Promise.any([
+        Promise.resolve(fetchImpl(url, options)),
+        promise.then(async () => null)
+      ]).finally(cancel);
```
SferaDev, line 78:
```suggestion
const response = await Promise.race([fetchImpl(url, options), promise.then(() => null)]).finally(cancel);
```

packages/client/src/util/fetch.ts
```diff
@@ -2,7 +2,13 @@ import { parseNumber, timeout, timeoutWithCancel } from './lang';
...
 const REQUEST_TIMEOUT = 5 * 60 * 1000; // 5 minutes
 
-export type RequestInit = { body?: any; headers?: Record<string, string>; method?: string; signal?: any };
+export type RequestInit = {
+  body?: any;
+  headers?: Record<string, string>;
+  method?: string;
+  signal?: any;
+  duplex?: 'half' | 'full';
+};
```
SferaDev, line 11:
```suggestion
};

```

packages/client/src/api/fetcher.ts
```diff
@@ -185,7 +185,10 @@ export async function fetch<
         method: method.toUpperCase(),
         body: await parseBody(body, headers),
         headers,
-        signal
+        signal,
+        // required in order to stream some body types to Cloudflare
+        // currently only supported in Node.js, we may have to feature detect this
+        duplex: 'half'
```
SferaDev, line 191:
```suggestion
        signal
```

packages/client/src/util/fetch.ts
```diff
@@ -2,7 +2,14 @@ import { parseNumber, timeout, timeoutWithCancel } from './lang';
...
```
SferaDev, line 11:
```suggestion
export type RequestInit = { body?: any; headers?: Record<string, string>; method?: string; signal?: any };
```

paulaguijarro (approved):

PR #1166, opened 2023-09-12
Fix Blob errors on Node 16

SferaDev:


eemmiillyy (approved):
LGTM!

.changeset/thirty-scissors-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
```
eemmiillyy, line 1: How does this file get used? I'm not familiar with the @changesets package.

SferaDev: I'll show you in our 1:1 today

PR #1173, opened 2023-09-13
feat: dynamic branch on init if using default `main` which does not exist

### LINKED ISSUE #1095: init defaults to the main branch
`init` defaults to using the `main` branch.

In case a database does not have the `main` branch and there is no env variable XATA_BRANCH pointing to an existing branch, init errors.

```
xata init        
🦋 Initializing project... We will ask you some questions.

✔ Select a workspace › ws
✔ Select a database or create a new one › mydb
✔ Do you want to use code generation in your project? › Generate TypeScript code
✔ Choose the output file for the code generator … src/xata.ts
 ›   Warning: Could not detect a package manager. Please install the @xata.io/client package manually.
i We are going to update your .env file to store an API key.
✔ Do you want to create a .gitignore file and ignore the .env file? … yes
i Added .env to .gitignore
    Error: database branch [mydb:main] not found
```

Let's explore possibilities such as:

- list available branches and let the user select one
- support a branch input param for the cli (as setting XATA_BRANCH isn't a very obvious way/needs research in docs)

eemmiillyy:
This PR adds the ability for the user to interactively select the branch they want to use if they are using the default `main` and that does not actually exist (during `xata init`)

If there are no branches for a database an error will be returned (not sure how one would get to this state).

Closes #1095

### DIFF
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -346,13 +346,24 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
     workspace: string,
     region: string,
     database: string,
-    options: { allowEmpty?: boolean; allowCreate?: boolean; title?: string } = {}
+    options: {
+      allowEmpty?: boolean;
+      allowCreate?: boolean;
+      title?: string;
+      checkDefaultExists?: { branch: string };
+    } = {}
   ): Promise<string> {
     const xata = await this.getXataClient();
     const { branches = [] } = await xata.api.branches.getBranchList({ workspace, region, database });
 
     const EMPTY_CHOICE = '$empty';
     const CREATE_CHOICE = '$create';
+
+    if (options.checkDefaultExists) {
+      if (branches.map(({ name }) => name).includes(options.checkDefaultExists.branch)) {
+        return options.checkDefaultExists.branch;
+      }
+    }
 
     if (branches.length > 0) {
       const choices = branches.map((db) => ({
--- a/cli/src/base.ts
+++ b/cli/src/base.ts
@@ -350,7 +350,8 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
       allowEmpty?: boolean;
       allowCreate?: boolean;
       title?: string;
-      useBranchIfExists?: { branch: string };
+      // Branch to default if exists
+      defaultBranch?: string;
     } = {}
   ): Promise<string> {
     const xata = await this.getXataClient();
@@ -359,10 +360,8 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
     const EMPTY_CHOICE = '$empty';
     const CREATE_CHOICE = '$create';
 
-    if (options.useBranchIfExists) {
-      if (branches.map(({ name }) => name).includes(options.useBranchIfExists.branch)) {
-        return options.useBranchIfExists.branch;
-      }
+    if (options.defaultBranch && branches.map(({ name }) => name).includes(options.defaultBranch)) {
+      return options.defaultBranch;
     }
 
     if (branches.length > 0) {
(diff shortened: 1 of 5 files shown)


cli/src/commands/init/index.ts
```diff
@@ -375,7 +388,7 @@ export default class Init extends BaseCommand<typeof Init> {
     return envFile;
   }
 
-  async writeEnvFile(workspace: string, region: string, database: string) {
+  async writeEnvFile(workspace: string, region: string, database: string, branch: string) {
```
eemmiillyy, line 391: The selected branch will be written into the env file

cli/src/commands/init/index.ts
```diff
@@ -157,14 +171,13 @@ export default class Init extends BaseCommand<typeof Init> {
...
 
-    const branch = this.getCurrentBranchName();
     if (schema) {
       await this.deploySchema(workspace, region, database, branch, schema);
     }
 
     // Run pull to retrieve remote migrations, remove any local migrations, and generate code
     await Pull.run([branch, '-f', '--skip-code-generation']);
-    await Codegen.runIfConfigured(this.projectConfig);
+    await Codegen.runIfConfigured(this.projectConfig, [`--branch=${branch}`]);
```
eemmiillyy, line 180: I left out the `--inject-branch` flag, we didn't write `main` into the output here so I didn't write the selected branch in their either.

SferaDev:
I'll review the code this afternoon, @kostasb can you give a functional review since you opened the issue?

kostasb:
Functionally works well!

cli/src/base.ts
```diff
@@ -346,14 +346,25 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
     workspace: string,
     region: string,
     database: string,
-    options: { allowEmpty?: boolean; allowCreate?: boolean; title?: string } = {}
+    options: {
+      allowEmpty?: boolean;
+      allowCreate?: boolean;
+      title?: string;
+      useBranchIfExists?: { branch: string };
```
SferaDev, line 353: What about:

```suggestion
      // Branch to default if exists
      defaultBranch?: string;
```

cli/src/base.ts
```diff
@@ -346,14 +346,25 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
...
 
     const EMPTY_CHOICE = '$empty';
     const CREATE_CHOICE = '$create';
 
+    if (options.useBranchIfExists) {
+      if (branches.map(({ name }) => name).includes(options.useBranchIfExists.branch)) {
+        return options.useBranchIfExists.branch;
+      }
+    }
+
```
SferaDev, line 367: Can be probably simplified

```suggestion
    if (branches.map(({ name }) => name).includes(options.defaultBranch)) {
      return options.useBranchIfExists.branch;
    }

```

cli/src/commands/init/index.ts
```diff
@@ -45,6 +45,8 @@ const packageManagers = {
   }
 };
 
+const defaultBranch = 'main';
```
SferaDev, line 48: [nit]

```suggestion
const DEFAULT_BRANCH = 'main';
```

cli/src/commands/init/index.ts
```diff
@@ -134,6 +136,18 @@ export default class Init extends BaseCommand<typeof Init> {
...
+    if (defaultBranch === branch) {
+      branch = await this.getBranch(workspace, region, database, {
+        allowCreate: false,
+        allowEmpty: false,
+        useBranchIfExists: {
+          branch: defaultBranch
+        }
+      });
+    }
+
```
SferaDev, line 150: [style] We can probably use const and avoid having a variable

SferaDev (approved):
The PR looks great, left some minor comments. Good work!

cli/src/base.ts
eemmiillyy: It's potentially undefined so I need to check it exists first. I can simplify it to this though:

```js
 if (options.defaultBranch && branches.map(({ name }) => name).includes(options.defaultBranch)) {
      return options.defaultBranch;
    }
```

SferaDev:
Merge it whenever you want. Thanks for the changes 😉

PR #1190, opened 2023-09-25
feat: import files csv

eemmiillyy:
This PR adds the ability to import remote and local files via CSV with only the path/URL specified in the CSV file. 

Closes #2332

TODO
- [x] Test`data:` as the input
- [x] Fix CORS issue for use in UI
- [x] Fix static import issue with fs in UI


packages/importer/src/columns.ts
```diff
@@ -149,23 +153,80 @@ export const coerceValue = (
...
+        isError: false
+      };
+    }
+    case 'file[]': {
+      // Cannot import multiple raw Base64 files because we will split in the wrong place
+      const promises = (value as string).split(/[,;|]/).map((uri) => parseFile(uri, proxyFunction));
+      const files = await Promise.all(promises);
+      const isError = files.some((file) => file === null);
+      const formatted = files.map(
+        (file) => ({ name: 'upload', mediaType: file?.mediaType, base64Content: file?.base64Content } as XataFile)
```
eemmiillyy, line 170: Just setting the name as 'upload' here since I don't think there's a reliable way to get the name? We also error in the backend when the file name has invalid characters.

packages/importer/src/columns.ts
```diff
@@ -149,23 +153,80 @@ export const coerceValue = (
...
+      const res = await parseFile((value as string).trim(), proxyFunction);
+      if (!res) return { value: null, isError: true };
+      return {
+        value: { name: 'upload', mediaType: res.mediaType, base64Content: res.base64Content } as XataFile,
+        isError: false
+      };
+    }
+    case 'file[]': {
+      // Cannot import multiple raw Base64 files because we will split in the wrong place
+      const promises = (value as string).split(/[,;|]/).map((uri) => parseFile(uri, proxyFunction));
```
eemmiillyy, line 166: This is a limitation with the current code right now. Base64 values will have these characters in them so the split will be wrong if the user intends multiple files

packages/importer/src/columns.ts
```diff
@@ -149,23 +153,80 @@ export const coerceValue = (
...
+  try {
+    if (uri.startsWith('data:')) {
+      const [mediaType, base64Content] = uri.split(',');
+      return new XataFile({ base64Content, mediaType });
+    } else if (uri.startsWith('http://') || uri.startsWith('https://')) {
+      const validUrl = new URL(uri);
+      if (proxyFunction) {
+        const response = await proxyFunction(validUrl.href);
+        if (!response) return null;
+        return XataFile.fromBase64(response.base64Content, { mediaType: response.mediaType });
```
eemmiillyy, line 191: I suppose we could change the type here and have the proxyFunction return Blob, but I couldn't get it to work (and I think the mime type would be lost) so I switched the type to make the proxy return the base64 content.

packages/importer/src/utils/lang.ts
```diff
@@ -1,7 +1,32 @@
...
+  return value !== null && value !== undefined;
+}
+
+export function compact<T>(arr: Array<T | null | undefined>): T[] {
+  return arr.filter(notEmpty);
+}
+
+export function isXataFile(value: unknown): value is XataFile {
+  // TODO fix this predicate
+  return isObject(value) && (value as any).mediaType !== undefined;
```
eemmiillyy, line 21: `value instanceof XataFile` was not returning true in situations where it should have

packages/importer/test/columns.test.ts
```diff
@@ -294,6 +305,96 @@ const coerceTestCases: { input: unknown; type: Schemas.Column['type']; options?:
       type: 'datetime',
       expected: { value: new Date('2020-01-01T00:00:00+01:00'), isError: false }
     },
+    // This test should have the response mocked
```
eemmiillyy, line 308: This should mock the fetch response so that we don't need a stable remote URL to check against

packages/importer/src/columns.ts
```diff
@@ -1,8 +1,9 @@
-import type { Schemas } from '@xata.io/client';
+import { Schemas, XataFile } from '@xata.io/client';
+
 import CSV from 'papaparse';
 import AnyDateParser from 'any-date-parser';
-import { ColumnOptions, ToBoolean } from './types';
-import { isDefined } from './utils/lang';
+import { ColumnOptions, ProxyFunction, ToBoolean } from './types';
```
SferaDev, line 5:
```suggestion
import { ColumnOptions, ToBoolean } from './types';
```

packages/importer/src/columns.ts
```diff
@@ -149,23 +153,67 @@ export const coerceValue = (
...
     }
+    case 'file': {
+      const file = await parseFile((value as string).trim(), proxyFunction);
+      if (!file) return { value: null, isError: true };
+
+      return { value: file, isError: false };
+    }
+    case 'file[]': {
+      // Limitation: Cannot import multiple raw Base64 files because we will split them in the wrong place
+      const items = (value as string).startsWith('data:') ? [value as string] : (value as string).split(/[,;|]/);
```
SferaDev, line 164: I think we can do better here

packages/importer/src/columns.ts
```diff
@@ -149,23 +153,67 @@ export const coerceValue = (
...
+const fetchFile = async (url: string) => {
+  const response = await fetch(url);
+  return await response.blob();
+};
+
+const parseFile = async (url: string, request = fetchFile): Promise<XataFile | null> => {
+  const uri = url.trim();
+  try {
+    if (uri.startsWith('data:')) {
+      const [mediaType, base64Content] = uri.split(',');
```
SferaDev, line 186: I think this doesn't work, will need to add a test

packages/importer/src/importer.ts
```diff
@@ -20,31 +19,78 @@ export const importBatch = async (
...
-    const result = await branchTransaction({ ...pluginOptions, pathParams, body: { operations } });
-    return { successful: result, errors };
+    const { results } = await branchTransaction({
+      ...pluginOptions,
+      pathParams: {
+        workspace: location.workspace,
+        region: location.region,
+        dbBranchName: `${location.database}:${location.branch}`,
+        // @ts-expect-error For some reason, database is required...
+        database: location.database
```
SferaDev, line 31: Found the root cause for this, will fix too

packages/importer/src/importer.ts
```diff
@@ -20,31 +19,78 @@ export const importBatch = async (
...
+    for (const [columnName, value] of Object.entries(row)) {
+      const files = Array.isArray(value) ? value : [value];
+      for (const file of files) {
+        try {
+          await putFile({
+            ...pluginOptions,
+            pathParams: {
+              workspace: workspace,
+              // @ts-expect-error For some reason we need to send it
+              database: database,
```
SferaDev, line 80: Same as before, found the root cause of this, will fix

SferaDev (approved):
All good! Let's merge it tomorrow in any case, I want to play with it a little bit on the frontend first

PR #1192, opened 2023-09-25
Increase pagination limits

tsg:
I've merged earlier today https://github.com/xataio/xata/pull/3117 which significantly increases the pagination limits:

* max page size from 200 to 1000
* max offset from 800 to 49000 (for a total of 50k)

e2e tests failed on staging and it makes sense because the client / e2e tests expected the old limits. I think this should fix it, but I didn't try it yet. I'll try running the tests against staging.


tsg:
Tested from local against staging and that passed. I'm not sure about the merge order, though? 

We force-merge this, then retry the deployment in the xata repo?

SferaDev:
Sounds good

SferaDev (approved):

PR #1205, opened 2023-10-02
feat: media type override

### LINKED ISSUE #1174: Optional property for file type
**Enhancement request**:
Add an optional property to set the file type (mediaType) with the binary file API upload call.

This will help with files as ArrayBuffer where it's not possible for us to guess the type, or in any other case where the user prefers to set the file type explicitly.

eemmiillyy:
This PR adds the ability to override the mediaType on the binary file upload call API 

Resolves #1174

### DIFF
--- a/packages/client/src/files/index.ts
+++ b/packages/client/src/files/index.ts
@@ -12,7 +12,8 @@ export type FilesPluginResult<Schemas extends Record<string, BaseData>> = {
   download: <Tables extends StringKeys<Schemas>>(location: DownloadDestination<Schemas, Tables>) => Promise<Blob>;
   upload: <Tables extends StringKeys<Schemas>>(
     location: UploadDestination<Schemas, Tables>,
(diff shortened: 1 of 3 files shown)


packages/client/src/files/index.ts
```diff
@@ -12,7 +12,8 @@ export type FilesPluginResult<Schemas extends Record<string, BaseData>> = {
   download: <Tables extends StringKeys<Schemas>>(location: DownloadDestination<Schemas, Tables>) => Promise<Blob>;
   upload: <Tables extends StringKeys<Schemas>>(
     location: UploadDestination<Schemas, Tables>,
-    file: BinaryFile
+    file: BinaryFile,
+    mediaType?: string
```
SferaDev, line 16:
```suggestion
    options?: { mediaType?: string }
```

SferaDev (approved):

PR #1206, opened 2023-10-02
feat: guess data uri columns

eemmiillyy:
This PR adds column guessing for datauri inputs as either `file` or `file[]` in the CSV importer

### DIFF
--- a/packages/importer/src/columns.ts
+++ b/packages/importer/src/columns.ts
@@ -5,6 +5,7 @@ import CSV from 'papaparse';
 import { ColumnOptions, ToBoolean } from './types';
 import { isValidEmail } from './utils/email';
 import { compact, isDefined } from './utils/lang';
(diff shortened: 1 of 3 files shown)


packages/importer/src/columns.ts
```diff
@@ -57,6 +57,8 @@ const isMultiple = <T>(value: T): boolean => isGuessableMultiple(value) || tryIs
 
 const isMaybeMultiple = <T>(value: T): boolean => isMultiple(value) || typeof value === 'string';
 
+const isDataUri = <T>(value: T): boolean => /(data:.*?;base64,.*?(?:[;,|]|$))/g.test(value as string);
```
SferaDev, line 60: Can we do an `isString(value) && //.test(value)` instead of the `as` casting?

SferaDev (approved):

PR #1208, opened 2023-10-04
Add support for importing vectors

SferaDev:


packages/importer/src/columns.ts
```diff
@@ -53,6 +53,24 @@ const parseMultiple = (value: string): string[] | null => {
...
+
+    const array = parseMultiple(String(value)) ?? [];
+    if (array.some((item) => !isFloat(item))) return null;
+
+    return array.length;
+  });
+
+  // Some old papers used vectors of length 300, ideally they have more than 200
+  const length = checks.find((value) => value !== null);
+  if (!length || length < 200) return false;
```
SferaDev, line 69:
```suggestion
  if (!length || length < 50) return false;
```

packages/importer/src/columns.ts
```diff
@@ -53,6 +53,24 @@ const parseMultiple = (value: string): string[] | null => {
...
+
+  return checks.every((value) => value !== null && value === length);
```
SferaDev, line 71:
```suggestion
  return checks.every((value) => value !== null);
```

packages/importer/src/columns.ts
```diff
@@ -53,6 +53,24 @@ const parseMultiple = (value: string): string[] | null => {
...
+    if (!isMultiple) return null;
+
+    const array = parseMultiple(String(value)) ?? [];
+    if (array.some((item) => !isFloat(item))) return null;
+
+    return array.length;
+  });
+
+  // Some old papers used vectors of length 300, ideally they have more than 200
+  const length = checks.find((value) => value !== null);
```
SferaDev, line 68:
```suggestion
  // Some old papers used vectors of length 300, ideally they have more than 200
```

packages/importer/src/parsers/jsonParser.ts
```diff
@@ -2,19 +2,34 @@ import JSON from 'json5';
...
 export const parseJson = async (options: ParseJsonOptions, startIndex = 0): Promise<ParseResults> => {
   const { data: input, columns: externalColumns, limit } = options;
 
   const array = Array.isArray(input) ? input : isObject(input) ? [input] : JSON.parse(input);
 
   const arrayUpToLimit = isDefined(limit) ? array.slice(0, limit) : array;
-  const columns = externalColumns ?? guessColumns(arrayUpToLimit, options);
-  const item = await coerceRows(arrayUpToLimit, columns, options);
+  const columnsGuessed = externalColumns ?? guessColumns(arrayUpToLimit, options);
+  const item = await coerceRows(arrayUpToLimit, columnsGuessed, options);
```
SferaDev, line 32: Add an error if they're not the same size as column

packages/importer/test/columns.test.ts
```diff
@@ -454,80 +498,279 @@ const coerceTestCases: { input: unknown; type: Schemas.Column['type']; options?:
...
+];
 
 describe('coerceValue', () => {
-  for (const { input, type, options, expected } of coerceTestCases) {
+  for (const { input, type, options, expected, extraColumn } of coerceTestCases) {
+    const column: Schemas.Column = { name: 'column', type, ...extraColumn };
     test(`coerceValue ${JSON.stringify(input)} should return ${JSON.stringify(expected)}`, async () => {
-      const result = await coerceValue(input, type, options);
+      const result = await coerceValue(input, column, options);
+      console.log(result);
```
eemmiillyy, line 773: Can probably remove the console.log

eemmiillyy (approved):

packages/importer/src/columns.ts
```diff
@@ -53,6 +53,20 @@ const parseMultiple = (value: string): string[] | null => {
...
+    const isMultiple = isGuessableMultiple(value);
+    if (!isMultiple) return null;
+
+    const array = parseMultiple(String(value)) ?? [];
+    if (array.some((item) => !isFloat(item))) return null;
+
+    return array.length;
+  });
+
+  return checks.every((length) => length !== null && length > 50);
```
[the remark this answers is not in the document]
SferaDev: No, but below 50 kind of doesn't make sense and it might be a false positive of a multiple or array (when we support them via postgres)

packages/importer/test/columns.test.ts
SferaDev, line 773:
```suggestion
```

PR #1212, opened 2023-10-06
feat: compatibility endpoint

### LINKED ISSUE #1207: Check for compatibility semver version
1. In the `base` command query the GitHub releases for the latest CLI one
2. Check if it's semver compatible and if not error out
3. Check if it's the latest, if it's not then show an info message to update to latest

eemmiillyy:
This PR checks the current CLI version running. If it is not semver compatible with our list of versions then an error is thrown. If it is not the latest version then we show a warning that a new version is available.

Resolves  #1207


cli/package.json
```diff
@@ -46,6 +47,7 @@
     "edge-runtime": "^2.5.3",
     "enquirer": "^2.4.1",
     "env-editor": "^1.1.0",
+    "http-call": "^5.3.0",
```
SferaDev, line 50: Can we use fetch? We support Node 18+ so should be no problem.

cli/src/hooks/init/compatibility.ts
```diff
@@ -0,0 +1,94 @@
...
+        `✨ A newer version of the Xata ${pkg.toUpperCase()} is now available: ${
+          tag.latest[pkg]
+        }. You are currently using version: ${currentVersion}`
+      );
+    // Compatibility check
+    const compatibility: Compatibility = JSON.parse(await readFile(compatibilityFile, 'utf-8'));
+    const compatibleRange = compatibility[pkg].compatibility
+      .filter((v) => v.compatible)
+      .map((v) => v.range)
+      .join('||');
```
eemmiillyy, line 46: There doesn't seem to be a utility for getting rid of this join, even a Range object expects a || separated string

.github/workflows/release.yml
```diff
@@ -47,10 +47,18 @@ jobs:
         run: pnpm build
 
       - name: Create Release Pull Request or Publish to npm
+        id: release
         uses: changesets/action@v1
         with:
           title: Release tracking
           publish: npx changeset publish
```
SferaDev, line 54:
```suggestion
          publish: npx changeset publish
          version: npx ../../scripts/changeset-version.mjs
```

And inside the mjs you run `changeset version` which gives you the version changes you need. So basically you use changesets CLI to do the version changes for you and append it to the compatibility.json file

compatibility.json
```diff
@@ -0,0 +1,18 @@
+{
+  "cli": {
+    "latest": "0.14.2",
+    "compatibility": [
+      {
+        "range": ">0.0.1"
```
eemmiillyy, line 6: I removed the `compatibility` field because I couldn't figure out how to validate those. I think it makes more sense to only pass valid ranges in this array and whatever does not satisfy those is invalid, WDYT? Could maybe even be simplified to just strings instead of objects?

cli/src/hooks/init/compatibility.ts
```diff
@@ -0,0 +1,109 @@
...
+export type PackageJson = { dependencies: Record<string, string> };
+
+export const check = async (params: {
+  pkg: 'cli' | 'sdk';
+  currentVersion: string;
+  compatibilityObj: Compatibility;
+}) => {
+  const { pkg, compatibilityObj } = params;
+  const currentVersion = semver.coerce(params.currentVersion)?.version as string;
+  const updateAvailable = semver.lt(currentVersion, compatibilityObj[pkg].latest);
```
eemmiillyy, line 25: If the user is using version ^1.1.1 and the newest version is 1.1.2 they will still see an update available log.

.github/workflows/release.yml
```diff
@@ -47,10 +47,12 @@ jobs:
         run: pnpm build
 
       - name: Create Release Pull Request or Publish to npm
+        id: release
```
SferaDev, line 50:
```suggestion
```

SferaDev (approved):

PR #1217, opened 2023-10-10
fix: summarize return types for json

### LINKED ISSUE #1215: Summarize API returns json column content as string
**Describe the bug**:
The `summarize` call outputs json column content as string instead of json.
Tested with client version: 0.26.7

**Repro**:
Where "myjson" is a column of type `json`:

```
const rec = await xata.db.games.read("rec_ck6m8s0tr08hl26rt2m0");
console.log(rec.myjson);

//response:
{
  "one": "content",
  "two": "content2"
}
```

```
const summary = await xata.db.games.summarize({
  columns: ["myjson"],
  summaries: {},
});

console.log(summary.summaries[0].myjson);

//response:
"{\"one\":\"content\",\"two\":\"content2\"}"
```

Reported on [Discord](https://discord.com/channels/996791218879086662/1159891484418510949/1159891484418510949).

eemmiillyy:
Resolves #1215

### DIFF
--- a/packages/client/src/schema/repository.ts
+++ b/packages/client/src/schema/repository.ts
@@ -1911,8 +1911,14 @@ export class RestRepository<Record extends XataRecord>
         },
         ...this.#getFetchProps()
       });
-
-      return result;
+      const schemaTables = await this.#getSchemaTables();
+      const response = {
+        ...result,
+        summaries: result.summaries.map((summary) =>
+          initObject(this.#db, schemaTables, this.#table, summary, data.columns ?? [])
+        )
+      };
+      return response;
     });
   }
 
@@ -2107,7 +2113,7 @@ export const initObject = <T>(
   }
 
   const record = { ...data };
-  const metadata =
+  const metadata: Object | undefined =
     xata !== undefined
       ? { ...xata, createdAt: new Date(xata.createdAt), updatedAt: new Date(xata.updatedAt) }
       : undefined;
@@ -2134,7 +2140,10 @@ export const initObject = <T>(
     return db[table].delete(record['id'] as string);
   };
 
-  record.xata = Object.freeze(metadata);
+  if (metadata) {
+    record.xata = Object.freeze(metadata);
+  }
+
   record.getMetadata = function () {
     return record.xata;
   };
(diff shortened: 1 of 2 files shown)


packages/client/src/schema/repository.ts
```diff
@@ -1911,8 +1911,14 @@ export class RestRepository<Record extends XataRecord>
...
-
-      return result;
+      const schemaTables = await this.#getSchemaTables();
+      const response = {
+        ...result,
+        summaries: result.summaries.map((summary) =>
+          initObject(this.#db, schemaTables, this.#table, summary, data.columns ?? [])
+        )
+      };
+      return response;
```
SferaDev, line 1921:
```suggestion
      return {
        ...result,
        summaries: result.summaries.map((summary) =>
          initObject(this.#db, schemaTables, this.#table, summary, data.columns ?? [])
        )
      };
```

packages/client/src/schema/repository.ts
```diff
@@ -2107,7 +2113,7 @@ export const initObject = <T>(
   }
 
   const record = { ...data };
-  const metadata =
+  const metadata: Object | undefined =
```
SferaDev, line 2116: Shouldn't be needed, right? (`Object` is normally something we avoid)

```suggestion
  const metadata =
```

SferaDev (approved):
Small comments, other than that looks great

packages/client/src/schema/repository.ts
eemmiillyy: It appears to be necessary, without it I get: `error  Unexpected any value in conditional. An explicit comparison or type cast is required  @typescript-eslint/strict-boolean-expressions`

packages/client/src/schema/repository.ts
```diff
@@ -2134,7 +2139,10 @@ export const initObject = <T>(
     return db[table].delete(record['id'] as string);
   };
 
-  record.xata = Object.freeze(metadata);
+  if (metadata !== undefined) {
```
SferaDev, line 2142:
```suggestion
  if (!!metadata) {
```

PR #1220, opened 2023-10-11
[Experimental] Support for prisma

SferaDev:


janpio:
👋 Heyho from Prisma. 

If you want, you can do a temporary PR similar to https://github.com/prisma/prisma-engines/pull/4342, which should make it simpler to run the Prisma test suites against the driver adapter. We are unfortunately not ready yet with a more general approach for that, but I am happy to work with getting that run in this temporary way.

janpio:
Actually, the driver living in your repo is the right spot - but _temporarily_ (via a PR) having a copy in our test setup can be useful for you I think. But we are currently also moving things around (soon the driver adapters live in `prisma/prisma`...), so maybe we wait a few days until we tackle that.

How do you think about transactions? As far I know your client does not support that yet. Many Prisma Client queries use transactions under the hood (at least as soon as the queries become a tiny bit complex). Would you want those to be not working, or are you ok skipping the transaction guarantees for your users?

SferaDev:
We are working to offer full postgres compatibility, including transactions in the near future. We will even expose a TCP wire protocol connection string and likely have two adapters HTTP and wire protocol. But all of that is blocked until we ship the full thing.

packages/plugin-client-prisma/src/driver.ts
```diff
@@ -14,65 +23,47 @@ abstract class XataQueryable implements Queryable {
...
-    this.#_tables = _tables ?? [];
   }
 
   override async performIO(query: Query): Promise<Result<PerformIOResult>> {
-    const { formatted, table } = prepareSql(query, this.#_tables);
-    const result = await this.client.sql({ statement: formatted, params: query.args });
-    const resultFormatted = { ...result, table };
-    return ok(resultFormatted);
+    const { sql, args: values } = query;
+    return ok(await this.xata.sql(sql, values, { arrayMode: true, fullResults: true }));
```
SferaDev, line 66:
```suggestion
    return ok(await this.xata.sql(sql, values));
```

packages/plugin-client-prisma/src/driver.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (warning) debug(`${tag} %O`, warning);
+
+    const [columnNames, columnTypes] = Object.fromEntries(columns as any).reduce(
+      ([names, types]: [string[], ColumnType[]], [name, { type_name }]: [string, { type_name: string }]) => {
+        names.push(name);
+        types.push(fieldToColumnType(type_name));
+        return [names, types];
+      },
+      [[], []] as [string[], ColumnType[]]
+    );
```
SferaDev, line 42: TODO: Improve this

Issue #1081, opened 2023-07-13
Support totalCount in search response


kostasb:
The search API now returns the total count of records matched, alongside the matched records ref. https://xata.io/docs/api-reference/db/db_branch_name/search#free-text-search

Clients can use this information to determine whether there are more results pages to fetch.

SferaDev:
I checked and we would need to do a breaking change to support totalCount in the SDK. We were planning some breaking changes with a v1, so adding it to the list

PR #1232, opened 2023-10-31
feat: add total count to search responses

eemmiillyy:
This PR adds the `totalCount` field to search responses (`search`,`search.byTable`, `search.all`, and `vectorSearch`)

It contains breaking changes to the search APIs as the response is no longer an array, but an object, for example:
```javascript
{ records: [], totalCount: 1 }
```
Note that: `search.byTable` returns a records object rather than an records array. Each table name is a key pointing to an array, for example:
```javascript
{ records: { users: [], posts: [] }, totalCount: 1 }
```

Resolves https://github.com/xataio/client-ts/issues/1081

### DIFF
--- a/test/integration/vectorSearch.test.ts
+++ b/test/integration/vectorSearch.test.ts
@@ -55,6 +55,7 @@ describe('search', () => {
   test('with size', async () => {
     const { records: results, totalCount } = await xata.db.users.vectorSearch('vector', [1, 2, 3, 4], { size: 2 });
 
+    // The total count may be higher than the size
     expect(totalCount).toEqual(4);
     expect(results.map((r) => r.full_name)).toEqual(['r4', 'r1']);
   });
--- a/test/integration/vectorSearch.test.ts
+++ b/test/integration/vectorSearch.test.ts
@@ -39,59 +39,69 @@ afterEach(async (ctx) => {
 
 describe('search', () => {
   test('search 1 2 3 4', async () => {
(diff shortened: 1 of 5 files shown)


test/integration/vectorSearch.test.ts
```diff
@@ -39,59 +39,69 @@ afterEach(async (ctx) => {
...
 
+    expect(totalCount).toEqual(4);
     expect(results.map((r) => r.full_name)).toEqual(['r2', 'r3', 'r4', 'r1']);
   });
 
   test('with size', async () => {
-    const results = await xata.db.users.vectorSearch('vector', [1, 2, 3, 4], { size: 2 });
+    const { records: results, totalCount } = await xata.db.users.vectorSearch('vector', [1, 2, 3, 4], { size: 2 });
 
+    expect(totalCount).toEqual(4);
```
eemmiillyy, line 58: Worth noting that the backend seems to return the total count of results regardless of the page size specified, so I matched against 4 here rather than 2

SferaDev: Please add a comment

SferaDev:
I'll fully review it on Thursday when I'm back. Only concern is in `byTable` what happens if I have a column called `totalCount`?

eemmiillyy:
@SferaDev Yeah that's a good point, I've now nested under `records: {...}` to avoid this edge case

SferaDev (approved):

PR #1243, opened 2023-11-08
fix: better branch create error

### LINKED ISSUE #1238: Gracefully handle gateway timeouts when creating branches
**Describe the bug**
When a branch create command encounters a timeout , the CLI shows:

`Error: Unexpected token < in JSON at position 0`

It should instead give a better report of the error.

eemmiillyy:
Resolves #1238

### DIFF
--- a/cli/src/commands/branch/create.ts
+++ b/cli/src/commands/branch/create.ts
@@ -31,12 +31,17 @@ export default class BranchCreate extends BaseCommand<typeof BranchCreate> {
 
     const { from } = flags;
 
-    const result = await xata.api.branches.createBranch({ workspace, region, database, branch, from });
+    try {
+      const result = await xata.api.branches.createBranch({ workspace, region, database, branch, from });
 
-    if (this.jsonEnabled()) return result;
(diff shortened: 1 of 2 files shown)


cli/src/commands/branch/create.ts
```diff
@@ -31,12 +31,17 @@ export default class BranchCreate extends BaseCommand<typeof BranchCreate> {
...
+    try {
+      const result = await xata.api.branches.createBranch({ workspace, region, database, branch, from });
+
+      if (this.jsonEnabled()) return result;
+
+      const message = `Branch ${branch} successfully created`;
+      this.success(message);
+    } catch (err) {
+      err instanceof Error
+        ? this.error(`${err.message.includes('Unexpected token < in JSON') ? 'Failed to create branch' : err.message}`)
```
eemmiillyy, line 43: Checking for the string that we know if thrown during a timeout. Could alternatively override all error messages from backend but I figured that some of them may actually be useful.

SferaDev (approved):

PR #1251, opened 2023-11-17
Remove Xata Workers

SferaDev:


cli/src/workers.ts
```diff
@@ -1,266 +0,0 @@
...
-              },
-              Function: {
-                enter(path) {
-                  if (isXataWorker(path)) {
-                    const args = (path.parent as CallExpression).arguments as any[];
-                    const workerName = args[0]?.value;
-                    if (!workerName || typeof workerName !== 'string') {
-                      console.error(`Found a worker without a name in file ${file}`);
-                    } else {
-                      functions[workerName] = path.toString();
```
eemmiillyy, line 127: Is `path` here the path to he user defined worker module which could be defined anywhere in their project, which will have its contents transpiled, bundled below, wrapped in this `workerCode` and then uploaded to CF? 

The user was able to define their own handlers that would be executed in CloudFlare at a certain endpoint right?

eemmiillyy (approved):

Issue #1256, opened 2023-11-22
CLI import csv errors with invalid UTF8 value beyond certain file size


kostasb:
**Describe the bug**
Beyond a certain CSV file size, depending on the data content, the xata CLI throws: Import Error "invalid UTF8 value".

Import command:

```
xata import csv sample.csv --create --table repro --columns=a,b,c --types=string,string,string
```

In the repro file, the max working file size is 65545 bytes. Adding one more character in the file, triggers the error.

Note, this does not affect the "Import a CSV file" in the UI, only the CLI. Tested with `@xata.io/cli/0.14.4`

PR #1257, opened 2023-11-22
CLI import: Added encoding flag

richardgill:
Fixes: https://github.com/xataio/client-ts/issues/1256

Some files were not being imported properly via the CLI.

Here is an example file: https://github.com/xataio/file-import-samples/blob/main/samples/encoding2.csv provided by @kostasb it contains lots of non-ascii characters.

To fix the issue I've defaulted the encoding node uses to read files to `utf8`. Users can also explicitly specify the encoding. 

## Testing

`xatadevbuild import csv ~/Workspace/xata/file-import-samples/generatedSamples/large1M.csv --create --table repro1` still works as expected.

`xatadevbuild import csv ~/Workspace/xata/file-import-samples/samples/encoding2.csv --create --table repro1` was broken -> now fixed.

### DIFF
--- a/.changeset/nervous-peas-think.md
+++ b/.changeset/nervous-peas-think.md
@@ -1,0 +1,5 @@
+---
+"@xata.io/cli": patch
+---
+
+CLI import: Added encoding flag, switch to utf8 by default. Fixes some obscure encoding issues.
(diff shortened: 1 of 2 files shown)


eemmiillyy (approved):
Nice

PR #1263, opened 2023-11-27
Don't attempt to retrieve `tables` on runtime

SferaDev:
First steps towards using the new schema endpoint for pgroll branches


eemmiillyy (approved):
Let's remove the check annotation warning?

PR #1268, opened 2023-11-27
Handle error from API on search e2e tests

eminano:
The aggregate endpoint can now return a 404 if the underlying opensearch schema doesn't exist. This can happen if the table is created and immediately queried, since it needs time to replicate through pgstream. 
This PR adds a try/catch to make sure errors returned from the API don't stop the indexing wait loop.

### DIFF
--- a/test/integration/aggregate.test.ts
+++ b/test/integration/aggregate.test.ts
@@ -365,7 +365,13 @@ async function waitForSearchIndexing(): Promise<void> {
       return waitForSearchIndexing();
     }
   } catch (error) {
-    await new Promise((resolve) => setTimeout(resolve, 2000));
-    return waitForSearchIndexing();
+    if (isHttpError(error) && !String(error.status).startsWith('4')) {
+      await new Promise((resolve) => setTimeout(resolve, 2000));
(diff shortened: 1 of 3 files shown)


test/integration/aggregate.test.ts
```diff
@@ -358,8 +358,13 @@ describe(
 );
 
 async function waitForSearchIndexing(): Promise<void> {
-  const { aggs } = await xata.db.teams.aggregate({ total: { count: '*' } });
-  if (aggs.total !== teams.length) {
+  try {
+    const { aggs } = await xata.db.teams.aggregate({ total: { count: '*' } });
+    if (aggs.total !== teams.length) {
```
SferaDev, line 363: If we negate the condition, we can move the setTimeout and the recursive call outside the try/catch and de-duplicate the retry code (the catch can be left empty with a //noop comment so that the linter doesn't complain).

```suggestion
    if (aggs.total === teams.length) {
      return;
```

SferaDev (approved):
Thanks! Looks good, added a very minor suggestion to avoid repeating the retry logic, but it's up to you to implement it.

PR #1281, opened 2023-12-14
Fix issue with default public access

SferaDev:
We were setting defaults and overriding the server side default on create. Making everything optional to avoid the problem.

### DIFF
--- a/packages/client/src/schema/files.ts
+++ b/packages/client/src/schema/files.ts
@@ -20,11 +20,11 @@ export class XataFile {
   /**
    * Name of the file.
    */
-  public name: string;
+  public name?: string;
   /**
    * Media type of the file.
    */
-  public mediaType: string;
+  public mediaType?: string;
   /**
    * Base64 encoded content of the file.
    */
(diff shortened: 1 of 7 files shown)


packages/client/src/schema/files.ts
```diff
@@ -20,56 +20,56 @@ export class XataFile {
   /**
    * Name of the file.
    */
-  public name: string;
+  public name?: string;
```
eemmiillyy, line 23: All of these being optional feels like kind of a weird API for the user. Maybe we should at least add JSDocs for the defaults if not specified (if there are any)?

eemmiillyy (approved):

packages/client/src/schema/files.ts
SferaDev: Yeah, it's true that it feels weird. However that's what happens with any column right now. We could remove the undefined bit in column selection. I'll open a ticket and once I clean-up the `object notation` deprecated paths I'll attempt to improve this too.

PR #1287, opened 2023-12-15
Make XataApiClient to use ES Proxies

SferaDev:


packages/client/src/api/client.ts
```diff
@@ -21,1907 +18,72 @@ export interface XataApiClientOptions {
...
+      return new Proxy(this, {
+        get: (_target, namespace: keyof typeof operationsByTag) => {
+          if (operationsByTag[namespace] === undefined) {
+            return undefined;
+          }
+
+          return new Proxy(
+            {},
+            {
+              get: (_target, operation: keyof (typeof operationsByTag)[keyof typeof operationsByTag]) => {
```
eemmiillyy, line 71: Does this intercept  all operations/methods inside each API class? I guess the fetch functions that these operations/methods call are going to be intercepted?

packages/client/src/api/client.ts
```diff
@@ -21,1907 +18,72 @@ export interface XataApiClientOptions {
...
-    database: Schemas.DBName;
-  }): Promise<void> {
-    return operationsByTag.databases.deleteDatabaseGithubSettings({
-      pathParams: { workspaceId: workspace, dbName: database },
-      ...this.extraProps
-    });
-  }
+      });
+    }
+  } as unknown as { new (options?: XataApiClientOptions): XataApiProxy };
```
eemmiillyy, line 87: Is the `new` part necessary? This syntax is confusing/new to me

packages/client/src/api/client.ts
```diff
@@ -21,1907 +18,72 @@ export interface XataApiClientOptions {
...
-      | {
-          table: string;
-          filter?: Schemas.FilterExpression;
-          target?: Schemas.TargetExpression;
-          boosters?: Schemas.BoosterExpression[];
+type UserProps = {
+  headers?: Record<string, unknown>;
+};
+
+type XataApiProxy = {
```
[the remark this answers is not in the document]
SferaDev: Yes, we need to rewrite the return type as the proxy won't infer the different properties/namespaces we will expose.

packages/client/src/api/client.ts
```diff
@@ -21,1907 +18,72 @@ export interface XataApiClientOptions {
...
+        apiKey,
+        trace,
+        clientName: options.clientName,
+        xataAgentExtra: options.xataAgentExtra,
+        clientID
+      };
+
+      return new Proxy(this, {
+        get: (_target, namespace: keyof typeof operationsByTag) => {
+          if (operationsByTag[namespace] === undefined) {
```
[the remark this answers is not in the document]
SferaDev: Exactly, this is a catch for runtime JS where a user tries to access a namespace that doesn't exist

packages/client/src/api/client.ts
SferaDev: The `get` here does not refer to HTTP `GET` operation. The `get` in a `Proxy` is trying to access a property/field in an object like `foo.bar`. In this particular case, in the class it's intercepting trying to read any member of it.

packages/client/src/api/client.ts
SferaDev: Haha, yeah! This is how `class` types are defined where the `new` clause refers to the constructor of a class. As a fun fact, classes are syntactic sugar for objects with a constructor that is injected in the new property.

packages/client/src/api/client.ts
```diff
@@ -1,14 +1,11 @@
 import { defaultTrace, TraceFunction } from '../schema/tracing';
 import { getAPIKey } from '../util/environment';
 import { FetchImpl, getFetchImplementation } from '../util/fetch';
+import { FlattenObject, RequiredKeys } from '../util/types';
```
SferaDev, line 4:
```suggestion
import { RequiredKeys } from '../util/types';
```

test/integration/smoke.test.ts
```diff
@@ -88,7 +88,7 @@ describe('API Client Integration Tests', () => {
         tableName: 'table'
       },
       body: {
-        record: { email: 'example@foo.bar' }
+        email: 'example@foo.bar'
```
eemmiillyy, line 91: We may have had a bug here and didn't notice. I don't think this part of the API was changed

SferaDev: I just added this test on Friday, weird I didn't catch it

eemmiillyy (approved):

PR #1295, opened 2023-12-19
Add workflow to publish next releases on npm

SferaDev:
- Publish pre releases to npm (see: https://github.com/xataio/client-ts/blob/next/.changeset/pre.json)
- Add a GitHub Action to attempt rebase main onto next (if there are conflicts we will need to manually rebase, but I think that's fine)
- Don't update compatibility json with next changes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - next
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,8 +47,15 @@ jobs:
       - name: Build
         run: pnpm build
 
+      - name: Update next channel
+        if: github.ref_name == 'next'
+        run: |
+          npx changeset version
+          npx changeset publish
+
       - name: Create Release Pull Request or Publish to npm
         uses: changesets/action@v1
+        if: github.ref_name == 'main'
         with:
           title: Release tracking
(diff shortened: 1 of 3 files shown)


scripts/changeset-version.mjs
```diff
@@ -6,6 +6,13 @@ import { execSync } from 'child_process';
 // Run "npx changeset version" to update the versions in the monorepo
 execSync('npx changeset version', { stdio: 'inherit' });
 
+// Check if we're on the main branch, if not, skip updating compatibility.json
+const branch = execSync('git rev-parse --abbrev-ref HEAD', { encoding: 'utf8' }).trim();
+if (branch !== 'main') {
+  console.log('Not on master, skipping compatibility.json update');
```
eemmiillyy, line 12:
```suggestion
  console.log('Not on main, skipping compatibility.json update');
```

.github/workflows/release.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - next
```
eemmiillyy, line 7: I think calling the next branch `dev` would also make sense, but both work

eemmiillyy (approved):

.github/workflows/release.yml
SferaDev: I want to strictly keep it as `next` major things go there, not things in development. So `dev` might be confusing, WDYT?

eemmiillyy: Okay makes sense. I just immediately thought of next.js haha but this sounds reasonable

PR #1302, opened 2023-12-19
fix: drizzle internal columns

eemmiillyy:
Resolves issue RE un queryable metadata with drizzle integration. 

Internal discussion: https://xata-hq.slack.com/archives/C057TM0BX1C/p1702977723000879

### DIFF
--- a/packages/plugin-client-drizzle/src/session.ts
+++ b/packages/plugin-client-drizzle/src/session.ts
@@ -78,8 +78,17 @@ export class XataPreparedQuery<T extends PreparedQueryConfig> extends PreparedQu
       );
 
     if (warning) console.warn(warning);
-
-    const rows = records.map((record) => fields.map((field) => record[field.path.join('.')]));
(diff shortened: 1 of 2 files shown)


packages/plugin-client-drizzle/src/session.ts
```diff
@@ -78,8 +78,17 @@ export class XataPreparedQuery<T extends PreparedQueryConfig> extends PreparedQu
       );
 
     if (warning) console.warn(warning);
-
-    const rows = records.map((record) => fields.map((field) => record[field.path.join('.')]));
+    const internalColumnNames = ['xata.version', 'xata.createdAt', 'xata.updatedAt', 'xata.deletedAt'];
```
eemmiillyy, line 81: This should maybe go somewhere else outside of this file?

SferaDev: These are changing right now with SQL. I think we can leave it as is and fix it for v1

SferaDev (approved):

PR #1304, opened 2023-12-19
Rewrite upload url path

SferaDev:
For large uploads we use direct urls to the service, we need to rewrite the domain in those endpoints. For now we are hard coding them, but we plan to add that information to the OpenAPI spec in the future.

The CodeQL warning can be ignored since this is local code and not server side checks when entering the variables.


packages/client/src/api/fetcher.ts
```diff
@@ -62,20 +63,45 @@ export type FetcherOptions<TBody, THeaders, TQueryParams, TPathParams> = {
...
+      switch (host) {
+        case 'production':
+          url = url.replace('xata.sh', 'upload.xata.sh');
+          break;
+        case 'staging':
+          url = url.replace('staging-xata.dev', 'upload.staging-xata.dev');
+          break;
+        case 'dev':
+          url = url.replace('dev-xata.dev', 'upload.dev-xata.dev');
+          break;
```
eemmiillyy, line 102: Might be nice to have an `exhaustiveCheck` in /client at some point.

eemmiillyy (approved):

PR #1317, opened 2024-01-09
Fix message in smoke test

SferaDev:


test/integration/smoke.test.ts
```diff
@@ -150,50 +150,33 @@ describe('API Client Integration Tests', () => {
...
       await api.branches.getBranchList({ workspace, database, region });
     }
   } catch (error) {
-    console.log(
-      `Waiting for create ${database === undefined ? 'API key' : 'database'} replication to finish... (${iteration})`
-    );
+    console.log(`Waiting for create ${database === undefined ? 'API key' : 'database'} replication to finish...`);
     await new Promise((resolve) => setTimeout(resolve, 2000));
-    return await waitForReplication(api, workspace, database, iteration + 1);
+    return await waitForReplication(api, workspace, database);
```
eemmiillyy, line 163: Is/was this going to loop forever if replication fails?

eemmiillyy (approved):
LGTM. Why do we need to remove the iteration logging in general though?

test/integration/smoke.test.ts
SferaDev: We have a timeout at 2 minutes that stops any loop.

SferaDev:
> Why do we need to remove the iteration logging in general though?

It wasn't adding more information and I only added 1/3. I had two options fix the 2 other loops or remove them altogether, chose to remove them to keep it simple.

test/integration/smoke.test.ts
eemmiillyy: Nice, TY!

PR #1323, opened 2024-01-19
feat: pgroll pull

eemmiillyy:
This PR adds support for pulling migrations for pgroll enabled branches. 

This is the flow for PgRoll enabled branches:

```
- pgroll history fetched
- are there existing migration files locally?
  - yes:  
    - are they in pgroll format?
      - no: delete the existing migrations and write response locally  
      - yes: compare to local and write missing migrations to disk
  - no: write new migrations to disk
```


.changeset/pre.json
```diff
@@ -1,5 +1,5 @@
 {
-  "mode": "pre",
+  "mode": "exit",
```
SferaDev, line 2: This should be `pre` still

eemmiillyy: For some reason the publish keeps failing without this: https://github.com/xataio/client-ts/actions/runs/7613023749/job/20731937567

cli/src/commands/diff/index.ts
```diff
@@ -34,24 +35,20 @@ export default class Diff extends BaseCommand<typeof Diff> {
     this.info(`Diff command is experimental, use with caution`);
 
     const localMigrationFiles = await getLocalMigrationFiles();
-    const schemaOperations = localMigrationFiles.flatMap((migrationFile) => migrationFile.operations);
+    // TODO decide what to do about this command for pgroll enabled branches and then fix type assertion
+    const schemaOperations = localMigrationFiles.flatMap(
+      (migrationFile) => (migrationFile as Schemas.MigrationObject).operations
```
SferaDev, line 40:
```suggestion
    const schemaOperations = localMigrationFiles.flatMap(
      // @ts-expects-error TODO decide what to do about this command for pgroll enabled branches and then fix type assertion
      (migrationFile) => migrationFile.operations
```

cli/src/migrations/files.ts
```diff
@@ -45,37 +47,46 @@ export async function getLocalMigrationFiles(): Promise<Schemas.MigrationObject[
     }
   }
 
-  const migrations: Schemas.MigrationObject[] = [];
+  const migrations: Schemas.MigrationObject[] | Schemas.PgRollMigrationHistoryItem[] = [];
```
SferaDev, line 50: In general I'd suggest:

```
(Schemas.MigrationObject | Schemas.PgRollMigrationHistoryItem)[]
```

.changeset/healthy-guests-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xata.io/cli': patch
+---
+
+Add support for pgroll enabled branches to `xata push`
```
SferaDev, line 5:
```suggestion
Add support for pgroll enabled branches to pull and push commands
```

SferaDev (approved):
I've been testing and no issues found in normal branches. With a pgroll branch I got into a state telling me there were both local and remote migrations, we will need to debug that.

cli/src/migrations/files.ts
```diff
@@ -29,7 +30,11 @@ async function readMigrationsDir() {
   }
 }
 
-export async function getLocalMigrationFiles(): Promise<Schemas.MigrationObject[]> {
+export type LocalMigrationFile =
+  | Schemas.MigrationObject
+  | (Schemas.PgRollMigrationHistoryItem & { id?: never; checksum?: never; operations?: never[] });
```
eemmiillyy, line 35: What does the `never` value do in this case? Isn't this the default for `Schemas.PgRollMigrationHistoryItem`?

SferaDev: So, in a union `A | B` you either have the properties of `A` or the properties of `B`. Trying to access `B.id` fails because `id` is not a property of `B`, however if we add an intersection of `never` (something that is never going to happen), `B.id` is now a property of `B` although you can never read or assign a value. In your code, it allowed to remove all the `as` casting trying to read `(A | B).id`.

eemmiillyy: I see, thank you!

PR #1328, opened 2024-01-24
Make sure sql function is called as a template

SferaDev:


.changeset/soft-ligers-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xata.io/client': patch
+---
+
+Fix untagged scenario
```
SferaDev, line 5:
```suggestion
Fix untagged scenario for SQL function
```

packages/client/src/sql/index.ts
```diff
@@ -1,28 +1,59 @@
...
+   * The SQL statement to execute.
+   * @example
+   * ```ts
+   * const { records } = await xata.sql<TeamsRecord>({
+   *  statement: `SELECT * FROM teams WHERE name = $1`,
+   *  params: ['A name']
+   * });
+   * ```
+   *
+   * Be careful when using this with user input and use prepared statements to avoid SQL injection.
```
SferaDev, line 17:
```suggestion
   * Be careful when using this with user input and use parametrized statements to avoid SQL injection.
```

test/integration/sql.test.ts
```diff
@@ -220,4 +220,26 @@ describe('SQL proxy', () => {
...
+    await expect(xata.sql<TeamsRecord>`SELECT * FROM teams ORDER BY name ${order}`).rejects.toThrow(
+      'invalid SQL: unused parameters: used 0 of 1 parameters'
+    );
+  });
+
+  test("calling xata.sql with invalid prepared statement doesn't throw an error when bypassing prepared statement protection", async () => {
+    const order = 'ASC';
+    const { records } = await xata.sql<TeamsRecord>({
+      statement: `SELECT * FROM teams ORDER BY name ${order}`
+    });
```
tsg, line 242: Could you also add a test showing sending prepared parameters?

test/integration/sql.test.ts
```diff
@@ -220,4 +220,26 @@ describe('SQL proxy', () => {
...
+  test("calling xata.sql as a function throws an error because it's not safe", async () => {
+    // @ts-expect-error - Testing invalid usage
+    await expect(xata.sql('SELECT * FROM teams')).rejects.toThrow(
+      'Invalid usage of `xata.sql`. Please use it as a tagged template or with an object.'
+    );
+  });
+
+  test('calling xata.sql with invalid prepared statement', async () => {
+    const order = 'ASC';
+    await expect(xata.sql<TeamsRecord>`SELECT * FROM teams ORDER BY name ${order}`).rejects.toThrow(
```
eemmiillyy, line 233: Why is this statement invalid?

eemmiillyy (approved):

test/integration/sql.test.ts
SferaDev: Postgres weird stuff 🥲😔

test/integration/sql.test.ts
SferaDev: We already had one test for that case in the file:

```
    const { records, warning, columns } = await xata.sql<TeamsRecord>({
      statement: `INSERT INTO teams (name) VALUES ($1) RETURNING *`,
      params: ['Team ships 2']
    });
```

PR #1345, opened 2024-02-02
Add github action to update pgroll spec

SferaDev:
Checks for changes every 3 hours on work days or manually.


eemmiillyy (approved):

.github/workflows/pgroll-spec.yml
```diff
@@ -0,0 +1,41 @@
...
+          version: 8
+
+      - name: Use Node.js ${{ steps.nvm.outputs.NVMRC }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ steps.nvm.outputs.NVMRC }}
+          cache: 'pnpm'
+
+      - name: Check for pgroll spec updates
+        run: cd packages/pgroll && pnpm generate
```
andrew-farries, line 30: I'm late, but does this step fail if there are no changes generated? 

Otherwise how does the action avoid creating a PR when nothing has changed?

SferaDev: The step from `peter-evans/create-pull-request@v4` avoids creating a PR if there are no changes on commit.

There's a couple of things I want to improve though:

- Run prettier before so that the action doesn't fail when the precommit hook removes the formatting differences
- Change the branch name so that the action updates the existing open PR

PR #1350, opened 2024-02-02
feat: pgroll migration format

eemmiillyy:


cli/src/migrations/schema.ts
```diff
@@ -48,16 +48,22 @@ export const migrationFile = z.object({
   operations: z.array(migrationOperation)
 }) satisfies z.ZodType<Schemas.MigrationObject>;
 
+export const pgRollMigrationHistoryObject = z.object({
+  name: z.string(),
+  operations: z.array(z.any())
+});
```
SferaDev, line 54: I'm fine with the `any` for now, do you think it makes sense to add `@xata.io/pgroll` as a dependency and import the generated zod schemas?

cli/src/commands/pull/pull.test.ts
```diff
@@ -27,6 +27,60 @@ const promptsMock = prompts as unknown as ReturnType<typeof vi.fn>;
...
+        },
+        table: 'test'
+      }
+    }
+  ]
+};
+
+const pgrollMigration1: Schemas.PgRollMigrationHistoryItem = {
+  done: false,
+  migration: `{"name": "mig_cmkjcdrj7c92neg7lnmg", "operations": [{"drop_column": {"down": "", "table": "tester", "column": "Firstname"}}]}`,
```
SferaDev, line 49: If I'm not mistaken these are the mocked files in file storage, right? Shouldn't these be JSON.stringified as if they were a result of a `xata pull`?

cli/src/migrations/schema.ts
```diff
@@ -48,16 +48,22 @@ export const migrationFile = z.object({
...
 
-export const pgRollMigrationHistoryObject = z.object({
-  name: z.string(),
-  operations: z.array(z.any())
-});
+export type LocalMigrationFilePgroll = {
+  [K in keyof Schemas.PgRollMigrationHistoryItem]: K extends 'migration'
+    ? z.infer<typeof pgRollMigrationHistoryObject>
+    : Schemas.PgRollMigrationHistoryItem[K];
+};
```
SferaDev, line 69: Nice work with a key remapping type!! Love to see them in the wild 🚀 

In this case, however, maybe we can simplify it though so it's more readable?

Haven't tried them in a real IDE but a couple of alternative suggestions:

```suggestion
export type LocalMigrationFilePgroll = Schemas.PgRollMigrationHistoryItem & {
  migration: z.infer<typeof pgRollMigrationHistoryObject>
};
```

```suggestion
export type LocalMigrationFilePgroll = Omit<Schemas.PgRollMigrationHistoryItem, 'migration'> & {
  migration: z.infer<typeof pgRollMigrationHistoryObject>
};
```

If you think the key remapped type is clear enough we can leave it as-is, up to you 😉

cli/src/migrations/files.ts
```diff
@@ -32,7 +32,7 @@ export async function readMigrationsDir() {
 
 export type LocalMigrationFile =
   | Schemas.MigrationObject
-  | (Schemas.PgRollMigrationHistoryItem & { id?: never; checksum?: never; operations?: never[] });
+  | (LocalMigrationFilePgroll & { id?: never; checksum?: never; operations?: never[] });
```
SferaDev, line 35: I find confusing that we have now 2 types called very similar `LocalMigrationFile` and `LocalMigrationFilePgroll`. I understand the naming but it could be difficult to follow in a few months and reason why we needed two types. Maybe they can be combined together in one or we find a different name. WDYT?

cli/src/migrations/schema.ts
eemmiillyy: I added it :)

cli/src/migrations/files.ts
eemmiillyy: Yeah agree, I tried to simplify it.

cli/src/commands/pull/pull.test.ts
eemmiillyy: These are the responses from the endpoint so `migration` field will be a string at this point. I'm reading the response and then parsing this into an object before saving to disk. 
When I read each file individually using this data, I call JSON.parse on the data first because that's the format it will have been written in.

SferaDev: Okay cool!

cli/src/commands/push/index.ts
```diff
@@ -101,17 +102,18 @@ export default class Push extends BaseCommand<typeof Push> {
...
       for (const migration of migrationsToPush) {
         try {
           await xata.api.branches.applyMigration({
             workspace,
             region,
             database,
             branch,
-            migration
+            // TODO remove assertion
+            migration: migration as any
```
SferaDev, line 116:
```suggestion
            // @ts-expect-error Backend API spec doesn't know all pgroll migrations yet
            migration
```

SferaDev (approved):

PR #1361, opened 2024-02-09
Add `toSerializable` to `getAll`

SferaDev:


packages/client/src/schema/pagination.ts
```diff
@@ -104,11 +104,50 @@ export function isCursorPaginationOptions(
...
+
+  toSerializable(): JSONData<Result>[] {
+    return JSON.parse(this.toString());
+  }
+
+  toString(): string {
+    return JSON.stringify(this.toArray());
+  }
+
+  map<U>(callbackfn: (value: Result, index: number, array: Result[]) => U, thisArg?: any): U[] {
```
eemmiillyy, line 140: What is `thisArg`?

packages/client/src/schema/pagination.ts
```diff
@@ -104,11 +104,50 @@ export function isCursorPaginationOptions(
 }
 
 export class RecordArray<Result extends XataRecord> extends Array<Result> {
+  constructor(overrideRecords?: Result[]);
+  constructor(...args: any[]) {
+    super(...RecordArray.parseConstructorParams(...args));
+  }
+
+  static parseConstructorParams(...args: any[]) {
+    // new <T>(arrayLength: number): T[]
```
eemmiillyy, line 113: What do the comments mean? 

Like someone could call `new RecordArray<SomeType>(1)`or `new RecordArray<SomeType>(["", ""])`?

packages/client/src/schema/query.ts
```diff
@@ -364,18 +365,18 @@ export class Query<Record extends XataRecord, Result extends XataRecord = Record
...
+    return array as unknown as PageRecordArray<Result>;
   }
 
   /**
    * Performs the query in the database and returns all the results.
    * Warning: If there are a large number of results, this method can have performance implications.
    * @returns An array of records from the database.
    */
-  getAll(): Promise<Result[]>;
+  getAll(): Promise<RecordArray<Result>>;
```
eemmiillyy, line 379: This is the main change that needed to happen right? What was the motivation for creating `PageRecordArray`?

test/integration/query.test.ts
```diff
@@ -86,6 +86,10 @@ describe('integration tests', () => {
 
     expect(teams).toHaveLength(1);
     expect(teams[0].name).toBe('Team fruits');
+
+    const serialized = teams.toSerializable();
```
eemmiillyy, line 90: Should we add a test for `toString` after this to make sure both serialisation and deserialisation work together?

eemmiillyy (approved):

packages/client/src/schema/pagination.ts
SferaDev: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#thisarg

packages/client/src/schema/pagination.ts
SferaDev: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Array#parameters

test/integration/query.test.ts
SferaDev: Sure

packages/client/src/schema/pagination.ts
eemmiillyy: Ah thanks. Never used that argument

PR #1364, opened 2024-02-12
Fix `highlight` not appearing in xata object

### LINKED ISSUE #1362: Add types for `highlight` object directly in search response
**Describe the bug**

The `xata` object for records returned using search API has still a standard `XataRecordMetadata` type:
```TS
type XataRecordMetadata = {
    /**
     * Number that is increased every time the record is updated.
     */
    version: number;
    /**
     * Timestamp when the record was created.
     */
    createdAt: Date;
    /**
     * Timestamp when the record was last updated.
     */
    updatedAt: Date;
};
```

There's no `highlight` or `score` options like in `SearchExtraProperties` type, although they are normally available in the code.

The only way to get it properly typed is to use the `getMetadata()` function on given record, however I can't see any reason why the `xata` object couldn't be typed directly.

**To Reproduce**

1. Call a search function on any table, e.g.:
```TS
const results = await xataClient.db.someTable.search(phrase, {
    target: [
      { column: "someColumn" },
    ],
    fuzziness: 2,
    prefix: "phrase",
    highlight: {
      enabled: true,
      encodeHTML: true,
    },
  })
```
2. Try to use `highlight` option later in the code:
```TSX
console.log(results.records[0].xata.highlight)
```
3. Observe a TS error:
```Shell
TS2339: Property highlight does not exist on type XataRecordMetadata
```

**Expected behavior**

TypeScript knows that `xata` object for records returned by `search` function contains the same properties as defined in `SearchExtraProperties` type:

```TS
type SearchExtraProperties = {
    table: string;
    highlight?: {
        [key: string]: string[] | {
            [key: string]: any;
        };
    };
    score?: number;
};
```

**Software version**

`"@xata.io/client": "^0.29.0"`

**Additional context**

SferaDev:
Fixes #1362


richardgill:
@SferaDev I remember that at one point we were removing `.xata` in favor of `.getMetadata()`. Can you remind me: Is this change consistent with the data parts of our SDK?

richardgill (approved):
LGTM

SferaDev:
> @SferaDev I remember that at one point we were removing `.xata` in favor of `.getMetadata()`. Can you remind me: Is this change consistent with the data parts of our SDK?

We're changing the behaviour for `.xata` object and likely removing both in v1. However let's fix the types meanwhile we release the breaking changes. Thanks @richardgill

PR #1371, opened 2024-02-13
Breaking changes to codegen

SferaDev:
- Make incremental builds the default setting
- Do not attempt to read the env variables by default
- Make the codegen more explicit
- Changes related to the new codegen file


packages/codegen/example/build-example.mjs
```diff
@@ -21,7 +21,7 @@ async function main() {
     language: 'typescript',
     existingCode: undoReplaceImport(typescriptContents)
   });
-  const { javascript: mjs, types } = await generate({ schema, databaseURL, language: 'javascript' });
+  const { javascript: mjs, types } = await generate({ schema, databaseURL, language: 'javascript', });
```
SferaDev, line 24:
```suggestion
  const { javascript: mjs, types } = await generate({ schema, databaseURL, language: 'javascript' });
```

packages/codegen/example/build-example.mjs
```diff
@@ -31,11 +31,11 @@ async function main() {
 }
 
 function replaceImport(source) {
-  return source.replaceAll('@xata.io/client', '../../client/src');
+  return source?.replaceAll('@xata.io/client', '../../client/src');
```
eemmiillyy, line 34: Is source possible undefined?

eemmiillyy (approved):
LGTM

packages/codegen/example/build-example.mjs
SferaDev: Yes, if the file doesn't exist previously

packages/codegen/src/codegen.ts
```diff
@@ -361,3 +301,27 @@ function emitDeclarations(code: string) {
 
   return files.get('index.d.ts');
 }
+
```
richardgill, line 304: @SferaDev What's the plan for cloudflare workers?

richardgill:
Looks like a good set of changes that moves us towards the notion doc 👍 🚀

richardgill (approved):

PR #1396, opened 2024-03-06
Update test to allow postgres

SferaDev:


exekias:
👍 does the test create other workspaces?

exekias (approved):

SferaDev:
> 👍 does the test create other workspaces?

IIRC this is the only place where we create databases that touch postgres features. We have some tests that create workspaces, but we don't test anything that requires pgroll migrations. If we see it fail, I'll open a new PR though.

PR #1412, opened 2024-03-15
fix: generate data native types

eemmiillyy:
Part of https://github.com/xataio/frontend-next/issues/3219

This PR uses the native Postgres types in order to seed random data for pgroll branches. It is still a best effort and will fail if there are constraints on fields such as length. The behaviour from non pgroll will not change much with this PR as the data seeded is the same except for reducing the default amount of characters seeded for strings.


packages/importer/src/random-data.ts
```diff
@@ -1,71 +1,75 @@
...
-    case 'multiple':
-      return faker.word.words(rand(1, 3)).split(' ');
-    case 'string':
-      return randomString(column.name);
-    case 'datetime':
+    case 'text':
+    case 'varchar':
+    case 'character varying':
+      if (columnCommentType === 'email') return faker.internet.email({ provider: 'acme.pets' });
+      if (column.type === 'link') return undefined;
```
eemmiillyy, line 48: I wonder if I should filter this out in the client as well, like I do with invalid links

packages/importer/src/random-data.ts
```diff
@@ -1,71 +1,75 @@
...
 import { Schemas } from '@xata.io/client';
+import { z } from 'zod';
 
 export function generateRandomData(table: Schemas.Table, size: number) {
   const records: Record<string, unknown>[] = [];
 
   for (let index = 0; index < size; index++) {
-    records.push(randomRecord(table.columns));
+    // TODO fix type
+    records.push(randomRecord(table.columns as any));
```
SferaDev, line 10: https://github.com/xataio/client-ts/pull/1416

SferaDev (approved):
LGTM. I wonder if just one word is to small for `text` though

packages/importer/src/random-data.ts
```diff
@@ -1,71 +1,75 @@
...
-      return faker.lorem.paragraphs(rand(2, 3));
-    case 'email':
-      return faker.internet.email({ provider: 'acme.pets' });
+function randomData(column: Schemas.Column & { comment?: string; pgType: string }) {
+  const columnCommentType = narrowStringType(column.comment);
+  // Note that this is a best effort and seeding may fail for invalid Xata columns
+  // that are foreign keys, or have constraints such as length attached to them.
+  switch (column.pgType) {
+    case 'boolean':
+    case 'bool':
```
SferaDev, line 31: With `pgType` feels like only one is needed

PR #1428, opened 2024-04-04
Polyfill `Buffer`

### LINKED ISSUE #1424: Buffer is not defined when deploying to Cloudflare
**Describe the bug**

Buffer is not defined when deploying to Cloudflare and using getFirst method
I don't think node:buffer is being used and so this error is risen

**To Reproduce**

**Expected behavior**

**Software version**

**Additional context**

SferaDev:
Some runtimes error out when `Buffer` is defined in user code, add a polyfill for it.

Closes #1424


packages/client/src/util/buffer.ts
```diff
@@ -0,0 +1,2292 @@
...
+  return str;
+}
+
+/**
+ * The encodings that are supported in both native and polyfilled `Buffer` instances.
+ */
+export type Encoding = 'ascii' | 'utf8' | 'utf16le' | 'ucs2' | 'binary' | 'hex' | 'latin1' | 'base64';
+
+// @ts-ignore
+if (typeof global !== 'undefined' && typeof global.Buffer !== 'undefined') Buffer = global.Buffer;
```
SferaDev, line 2290: Don't want to make it global

packages/client/src/util/buffer.ts
```diff
@@ -0,0 +1,2289 @@
+/* eslint-disable prefer-rest-params, prefer-spread */
+
+import * as base64 from './base64';
+
+export const K_MAX_LENGTH = 0x7fffffff;
+export const MAX_ARGUMENTS_LENGTH = 4096;
+
+class Buffer extends Uint8Array {
```
SferaDev, line 8:
```suggestion
// Based from https://github.com/feross/buffer
// Re-used some type checking from https://gist.github.com/baileyherbert/799c9277429497a40c5f28599494d46f
class Buffer extends Uint8Array {
```

richardgill (approved):
LGTM, thanks for the changes 👍

PR #1430, opened 2024-04-04
feat: pgroll cli edit

eemmiillyy:
Resolves https://github.com/xataio/client-ts/issues/1419

This PR makes the edit schema CLI command work with pgroll branches. Interactive source editing has been knowingly left out. The existing edit schema experience should not have really changed, this is mainly a refactor.  The generated migration that the user will be prompted to confirm or reject after submitting the migration will look different (pgroll migration format).


cli/src/commands/schema/edit.ts
```diff
@@ -294,593 +195,826 @@ Beware that this can lead to ${chalk.bold(
...
-            } else if (value) {
-              return 'The vectorDimension field must not be filled unless the type of the column is `vector`';
-            }
-            return true;
+          name: 'type',
+          message: `The type of the column ${xataTypes}`,
+          validate: (value: string) => {
+            if (value === undefined) return 'Type cannot be undefined';
+            if (emptyString(value)) return 'Type cannot be empty';
+            if (!xataTypes.includes(value))
```
eemmiillyy, line 506: Not allowing creation of types that are not "xata types"

cli/src/migrations/pgroll.ts
```diff
@@ -1,9 +1,10 @@
 import { Schemas, XataApiClient } from '@xata.io/client';
+import { migrationsDir, readMigrationsDir } from './files.js';
```
eemmiillyy, line 2: We should maybe have a shared util package that can be used in the front end as well? A lot of this is copied.

eemmiillyy:
I think we should maybe be calling adapt table after creating columns? If the user creates a table with the edit command, it won't have the internal ids. Should we automatically do that in the background? (@SferaDev)

Also, there is quite a bit of updating constraint logic in the front end that runs in asyncMigration if necessary. Should we also be doing that stuff inside a wrapper for `applyMigration` ?

cli/src/migrations/pgroll.ts
eemmiillyy: I created this since it should probably be a separate task https://github.com/xataio/client-ts/issues/1438

SferaDev:
@kostasb We'd really benefit from some dogfooding from your side if you have bandwidth 😄

kostasb:
Just noticing this, it's probably the same in the latest cli version and possibly out of scope (?), there is no option to remove tables or columns

eemmiillyy:
@kostasb you should be able to delete columns and tables with the delete key :)

kostasb:
> @kostasb you should be able to delete columns and tables with the delete key :)

That's great! Worth an issue to hint it within the CLI as it's a long shot for the user to discover

cli/src/commands/schema/edit.ts
```diff
@@ -1,107 +1,64 @@
...
+  'text',
+  'multiple',
+  'link',
+  'email',
+  'datetime',
+  'vector',
+  'json',
+  'file',
+  'file[]'
+];
```
SferaDev, line 58: Even if we don't expose more postgres types for now, let's stop using xata types and move to Postgres ones.

cli/src/commands/schema/edit.ts
```diff
@@ -116,175 +73,120 @@ export default class EditSchema extends BaseCommand<typeof EditSchema> {
...
         'Use the ↑ ↓ arrows to move across the schema, enter to edit or add things, delete or backspace to delete things.'
     });
+
+    for (const table of [
+      ...this.branchDetails!.schema.tables,
+      ...this.tableAdditions.map((addition) => ({
+        name: addition.name,
+        columns: []
+      }))
+    ]) {
```
SferaDev, line 117: We can probably move this to a new variable. Also the `!` looks fishy 😄

cli/src/commands/schema/edit.ts
SferaDev: We should revisit this in a follow-up

cli/src/commands/schema/edit.ts
```diff
@@ -294,593 +196,832 @@ Beware that this can lead to ${chalk.bold(
...
+  const isTableDeleted = (name: string) => {
+    return localTableDeletions.find(({ name: tableName }) => tableName === name);
+  };
+
+  // Remove column edits, additions and deletions for tables that are deleted
+  for (const [tableName, _] of Object.entries({
+    ...localColumnAdditions,
+    ...localColumnEdits,
+    ...localColumnDeletions
+  })) {
```
SferaDev, line 735:
```suggestion
  for (const tableName of Object.keys({
    ...localColumnAdditions,
    ...localColumnEdits,
    ...localColumnDeletions
  })) {
```

cli/src/commands/schema/edit.ts
```diff
@@ -294,593 +196,832 @@ Beware that this can lead to ${chalk.bold(
...
+        from: name,
+        to: newName
       }
-    }
+    };
+  });
+
+  const columnEdits: ({ alter_column: OpAlterColumn } | { drop_constraint: OpDropConstraint })[] = [];
+  for (const [_, columns] of Object.entries(localColumnEdits)) {
+    for (const [_, data] of Object.entries(columns)) {
```
SferaDev, line 876: We can use Object.values instead

cli/src/commands/schema/edit.ts
```diff
@@ -294,593 +196,832 @@ Beware that this can lead to ${chalk.bold(
...
+      const { name, nullable, unique, originalName } = data;
+      formatColumnDataToPgroll([data]).map(({ column: _, tableName }) => {
```
SferaDev, line 878:
```suggestion
      formatColumnDataToPgroll([data]).map(({ tableName }) => {
```

cli/src/commands/schema/edit.ts
```diff
@@ -294,593 +196,832 @@ Beware that this can lead to ${chalk.bold(
...
-    });
+  return [...columnDeletions, ...tableDeletions, ...tableAdditions, ...columnAdditions, ...columnEdits, ...tableEdits];
+};
 
-    this.success('Migration completed!');
-  }
-}
+const isReservedXataFieldName = (name: string) => {
+  return name.toLowerCase().startsWith('xata_');
+};
```
SferaDev, line 972: We shouldn't need this check anymore I think?

SferaDev (approved):

PR #1459, opened 2024-04-30
fix script

eemmiillyy:


SferaDev (approved):

.github/workflows/release-cli-assets.yml
```diff
@@ -3,6 +3,12 @@ name: Release CLI Assets
 on:
   workflow_call:
   workflow_dispatch:
+    inputs:
+      published-packages:
+        description: 'Published packages'
+        required: true
+        default: ''
```
SferaDev, line 10: Can we add a default that shows the expected format?

PR #1487, opened 2024-05-29
Add batch endpoint to SQL

SferaDev:


packages/client/src/sql/index.ts
```diff
@@ -31,6 +31,30 @@ export type SQLQueryParams<T = any[]> = {
...
+    statement: string;
+    /**
+     * The parameters to pass to the SQL statement.
+     */
+    params?: any[];
+  }[];
+  /**
+   * The consistency level to use when executing the queries.
+   */
+  consistency?: 'strong' | 'eventual';
```
eemmiillyy, line 51: What's the default? Should that info be included in the comment here? Same question for `responseType`

eemmiillyy (approved):

PR #1494, opened 2024-06-03
Make CLI backwards compatible with non pgroll branches

eemmiillyy:
PR to make CLI backwards compatible with non pgroll branches


cli/src/commands/import/csv.ts
```diff
@@ -80,6 +89,8 @@ export default class ImportCSV extends BaseCommand<typeof ImportCSV> {
     file: Args.string({ description: 'The file to be imported', required: true })
   };
 
+  pgrollEnabled = false;
```
SferaDev, line 92:
```suggestion
  #pgrollEnabled = false;
```

cli/src/commands/schema/edit.ts
```diff
@@ -115,775 +19,36 @@ export default class EditSchema extends BaseCommand<typeof EditSchema> {
...
+
+    const details = await getBranchDetailsWithPgRoll(xata, { workspace, region, database, branch });
+
+    if (isBranchPgRollEnabled(details)) {
+      const editNew = new EditSchemaNew(this.argv, config);
+      editNew.launch({
+        workspace,
+        region,
+        database,
+        branch
```
SferaDev, line 42: We can problably keep editnew in here instead of new file. And only have old one in a separate file.

cli/src/commands/schema/edit.ts
```diff
@@ -1,104 +1,8 @@
...
-        table: EditableTable;
-        column: EditableColumn;
-      };
-  message: string;
-  role?: string;
-  choices?: SelectChoice[];
-  disabled?: boolean;
-  hint?: string;
-};
+import EditSchemaOld from './editOld.js';
```
SferaDev, line 3: Proabbly better to follow conventions `edit-old`

SferaDev (approved):
Approved, can you do a final testing of everything and ask @kostasb to see if he can also do some testing? Just to be completely sure before merging.

kostasb:
Gave it a try, things work well.

My only commend is that for both pg and non-pg databases, xata init installed the client version 0.29.5. I think it would be best if it installed @next for pg branches ? Or do we plan to make the client figure out if it's a pg or non-pg branch?

eemmiillyy:
@kostasb yeah the CLI currently downloads the version the user has in their package.json. If they have none this is run `[packagemanager] install @xata.io/client` which I think just resolves to the latest version. 

For the case where there is no currently installed version of the SDK, the latest version will be downloaded and then this warning (PR [here](https://github.com/xataio/client-ts/pull/1507) ) would show prompting the user to manually change the version to next if on a pgroll branch. We could also try and pre-emptively download `next` for the user in this case. WDYT @SferaDev?

SferaDev:
Yeah, I think that at least prompting to update makes sense.

eemmiillyy:
So to summarise, for pgroll branches and xata init, if the user does not have a next version of SDK in package.json or has no version at all we will ask them Y/N if they want to download next branch (if no we just download latest)?

I can make a follow up PR to this one if we want to add that.

PR #1507, opened 2024-06-12
add hook for checking sdk version on codegen

eemmiillyy:


cli/src/hooks/next-check/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const getSdkVersion = async (): Promise<null | string> => {
+  const packageJson: any = JSON.parse(await readFile(`${path.join(process.cwd())}/package.json`, 'utf-8'));
+  return packageJson?.dependencies?.['@xata.io/client'] ? packageJson.dependencies['@xata.io/client'] : null;
+};
+
+const hook: Hook<'next-check'> = async function (options: Record<string, unknown>) {
+  if (options.pgrollEnabled) {
+    const sdkVersion = await getSdkVersion();
+    if (sdkVersion && sdkVersion !== 'next') {
```
eemmiillyy, line 13: This is what the dep would look like in package.json right? `"@xata.io/client": "next"`

Right now we have alpha versions (`0.0.0-nextv123`), so I could instead check for `-next.v` to cover both cases?

cli/src/hooks/next-check/index.ts
```diff
@@ -0,0 +1,21 @@
+import { Hook } from '@oclif/core';
+import { readFile } from 'fs/promises';
+import path from 'path';
+
+export const getSdkVersion = async (): Promise<null | string> => {
```
eemmiillyy, line 5: I duplicated this code from compatibility.ts. It doesn't seem to run when imported from another file, I haven't figured out why.

cli/src/hooks/next-check/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  const packageJson: any = JSON.parse(await readFile(`${process.cwd()}/package.json`, 'utf-8'));
+  return packageJson?.dependencies?.['@xata.io/client'] ? packageJson.dependencies['@xata.io/client'] : null;
+};
+
+const hook: Hook<'next-check'> = async function (options: Record<string, unknown>) {
+  if (options.pgrollEnabled) {
+    const sdkVersion = await getSdkVersion();
+    if (sdkVersion && sdkVersion !== 'next') {
+      this.warn(
+        "You are working with a pgroll enabled branch. We recommend using the 'next' version of the Xata SDK. (@xata.io/client@next)"
```
snide, line 15:
```suggestion
        "You are working with a Postgres enabled branch. We recommend using the 'next' version of the Xata SDK. (@xata.io/client@next)"
```

Most of the docs we focus on the Postgres nature of the toggle, not the pgroll side.

snide (commented):
Small suggestion.

cli/src/hooks/next-check/index.ts
eemmiillyy: Thank you!

cli/src/hooks/next-check/index.ts
```diff
@@ -0,0 +1,21 @@
+import { Hook } from '@oclif/core';
+import { readFile } from 'fs/promises';
+import path from 'path';
```
SferaDev, line 3:
```suggestion
```

cli/src/hooks/next-check/index.ts
SferaDev: Yeah, we should cover alpha versions. The `next` tag will never be there, exact versions are needed.

eemmiillyy: I added a check for `-next.` in the version

eemmiillyy:
@SferaDev this PR may no longer be necessary now that we decided to check and prompt here: https://github.com/xataio/client-ts/compare/feat/pgroll-next-prompt?expand=1

eemmiillyy:
Closing this in favor of https://github.com/xataio/client-ts/pull/1514

PR #1514, opened 2024-06-21
feat: pgroll next prompt

eemmiillyy:


cli/src/commands/init/index.ts
```diff
@@ -365,6 +390,23 @@ export default class Init extends BaseCommand<typeof Init> {
...
+      message: `You are working with a Postgres enabled branch. We recommend using the 'next' version of the Xata SDK. Do you want to install the recommended version (${pkg}@next)?`
+    });
+    if (installNextVersion) {
+      await this.runCommand(command, [...args.split(' '), `${pkg}@next`]);
+      this.log();
+    } else {
+      await this.runCommand(command, [...args.split(' '), `${pkg}`]);
+      this.warn(`Please install ${pkg}@next later`);
+    }
+    return Boolean(installNextVersion);
```
richardgill, line 407: More a philisophical question:

Should we be asking? Or should we just say something like:

"We installed @next because you're on a Postgres enabled branch"

richardgill (approved):
Left some comments / questions. 

But approving to unblock 👍

cli/src/commands/init/index.ts
eemmiillyy: I think it makes sense to download next if they're on a pgroll branch and there is no SDK version defined, but if they have a version defined we should always ask before overriding. I think it would be a relatively simple fix to this PR

richardgill:
Thanks for the changes, LGTM 👍

Issue #1541, opened 2024-07-22
Feature - Multi-version schema migration


divyenduz:
**Is your feature request related to a problem? Please describe.**

Xata UI app offers [multi-version schema migrations](https://xata.io/blog/multi-version-schema-migrations). 

**Describe the solution you'd like**

This functionality should also be offered in the Xata CLI

PR #1542, opened 2024-07-22
feat(multi-schema): add multi-schema capability to the CLI 

divyenduz:
Adds multi-version schema management to CLI by adding `xata migration <list, start, rollback, complete, status>` commands

Fix https://github.com/xataio/client-ts/issues/1541

## Done

- [x] base version of status, list, start, rollback, complete commands
- [x] Automated tests
- [x] Testing edge cases
  - [x] run status, list, rollback, complete on empty project
  - [x] run commands in unexpected states
- [x] Release change-set (and process)

## Pending

- [ ] Add history alias to list?

Demo/screenshots

## Start

Can be invoked with two variants

- `xata migration start <branch-name> migrationFile.json`
OR
- `xata migration start <branch-name> --migration-json '<inline-migration-operations>'`

Screenshot 2024-07-22 at 10 11 46

## Status

Prints the last migration job status

- `xata migration status <branch-name>`

Screenshot 2024-07-22 at 10 14 05

## List

Prints the migration history

- `xata migration list <branch-name>`

Screenshot 2024-07-22 at 10 15 34

## Rollback

- `xata migration rollback <branch-name>`

Screenshot 2024-07-22 at 10 16 34

## Complete

- `xata migration complete <branch-name>`

Screenshot 2024-07-22 at 10 17 27


cli/src/commands/migration/complete.ts
```diff
@@ -0,0 +1,68 @@
...
+      this.error(`No migration with ${migrationJobStatus.status} found. There is nothing to complete.`);
+    }
+
+    const migrationJob = await xata.api.migrations.completeMigration({
+      pathParams: {
+        ...commonParams
+      }
+    });
+
+    this.log(
```
divyenduz, line 61: I chose to print this message but I can also use `waitForMigrationToFinish` and automatically run status command after complete.

cli/src/commands/migration/rollback.ts
```diff
@@ -0,0 +1,68 @@
...
+      dbBranchName: `${database}:${branch}`
+    };
+
+    const migrationJobStatus = await xata.api.migrations.getBranchMigrationJobStatus({
+      pathParams: {
+        ...commonParams
+      }
+    });
+
+    const isActiveMigration = migrationJobStatus.status === 'completed' && migrationJobStatus.type === 'start';
```
divyenduz, line 50: Note: this is not 100% reliable, if for whatever reason a duplicate migration was submitted, the last migration status can be failed (see example)

`mig_job_cqdp2h97d1aiu327mf2g start failed failed 2024-07-20T10:28:22.45296Z a migration for schema "bb_8dd2korbmd52jfkbmbpfud6ass_d5mrcj" is already in progress`

Yet a migration can be active (the second last migration in this case will be active)

For this to be reliable, we need to ensure that UI, CLI can't call start a job when a migration is running AND any direct API access can't start it either.

That means, either backend needs to reject the job when a migration is already running or we need a new API to reliably detect a running migration instead of trying to find it from the last job status as an alternative we can use the migration history API and find any active migration.

cli/src/commands/migration/start.ts
```diff
@@ -0,0 +1,125 @@
...
+import { safeJSONParse, safeReadFile } from '../../utils/files.js';
+
+export default class MigrationStart extends BaseCommand<typeof MigrationStart> {
+  static description = 'Start a new migration';
+
+  static examples = [];
+
+  static migrationJsonFlag = {
+    'migration-json': Flags.string({
+      helpValue: `
```
divyenduz, line 15: TODO: need to check how this renders in CLI help, while currently hidden, I might have to turn this into a single line.

cli/src/commands/migration/status.ts
```diff
@@ -0,0 +1,81 @@
...
+      dbBranchName: `${database}:${branch}`
+    };
+
+    const migrationJobStatus = await xata.api.migrations.getBranchMigrationJobStatus({
+      pathParams: {
+        ...commonParams
+      }
+    });
+
+    const statusChalkColor = match(migrationJobStatus.status)
```
divyenduz, line 51: I like `ts-pattern` as it forces you to explicitly handle all states. Let me know if there are any reservations against introducing it to this repo.

cli/src/commands/migration/status.test.ts
```diff
@@ -0,0 +1,256 @@
...
+    };
+  } else {
+    return baseFetch(url, request);
+  }
+};
+
+promptsMock.mockReturnValue({ confirm: true, database: 'db1', workspace: 'test-1234' });
+
+describe('migration status', () => {
+  test('correctly detects if migration status is run in a project with no migrations', async () => {
```
eemmiillyy, line 206: Should we add a failed case? And maybe a test for non pgroll branches?

cli/src/commands/migration/status.test.ts
```diff
@@ -0,0 +1,256 @@
...
+        schema: { tables: [{ name: 'table1', columns: [{ name: 'description', type: 'string' }] }] }
+      })
+    };
+  }
+
+  if (url === `${baseUrl}/schema` && request.method === 'GET') {
+    return {
+      ok: true,
+      json: async () => ({
+        schema: {
```
eemmiillyy, line 59: Should we share this schema between the new tests? Seems to be the same/very similar

cli/src/commands/migration/start.ts
```diff
@@ -0,0 +1,125 @@
...
+    const xata = await this.getXataClient();
+    const { workspace, region, database, branch } = await this.getParsedDatabaseURLWithBranch(
+      flags.db,
+      args.branch,
+      true
+    );
+
+    const details = await getBranchDetailsWithPgRoll(xata, { workspace, region, database, branch });
+
+    if (!isBranchPgRollEnabled(details)) {
```
eemmiillyy, line 82: This is used in a few places so we could have a shared util/helper perhaps

cli/src/commands/migration/rollback.ts
eemmiillyy, line 50: Could also be a shared helper/util

cli/src/commands/migration/rollback.ts
```diff
@@ -0,0 +1,68 @@
...
+    branch: Args.string({ description: 'The branch to rollback the migration in', required: true })
+  };
+
+  static hidden = true;
+
+  async run() {
+    const { args, flags } = await this.parseCommand();
+
+    const xata = await this.getXataClient();
+    const { workspace, region, database, branch } = await this.getParsedDatabaseURLWithBranch(
```
eemmiillyy, line 26: What happens if an invalid or non existent branch is specified?

cli/src/commands/migration/list.ts
```diff
@@ -0,0 +1,72 @@
...
+        chalk.cyan(migration.name),
+        migration.migrationType,
+        migration.done ? chalk.green('complete') : chalk.yellow('in progress'),
+        chalk.gray(migration.parent),
+        migrationOperationsStr.substring(0, TRUNCATE_LIMIT) + chalk.gray(truncatedStr)
+      ];
+    });
+
+    this.printTable(tableHeaders, tableRows);
+    this.log();
```
eemmiillyy, line 70: Is the empty log adding a space?

cli/src/commands/migration/complete.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  });
+
+  test('correctly starts the migration complete job, if an active migration is found', async () => {
+    const config = await Config.load();
+    const command = new MigrationComplete(['main'], config);
+    const log = vi.spyOn(command, 'log');
+    fetchMock.mockImplementation(fetchRunningMigrationWithSuccessfulComplete);
+    await command.run();
+    expect(log).toHaveBeenCalledWith(
+      `Migration complete started with Job ID mig_job_cqfltm0knc8jbigtjb9g. Please use the xata migration status main command to check its status`
```
eemmiillyy, line 196: Should we add quotes around the command: `xata migrate status main` in the log?

divyenduz: Good idea, it was already using `chalk.gray` but quotes will make it clearer (and the tests more readable too). Done.

cli/src/commands/migration/list.ts
divyenduz: Yes, I noticed that this was making the text more readable by giving a space between the end of command text and the next prompt.

cli/src/commands/migration/rollback.ts
divyenduz: CLI prints: 

```bash
Error: database branch [localxb:404_branch] not found
```

As this endpoint returns 404 with data

```bash
$ curl -I -H 'Authorization: Bearer <API-KEY>' 'https://<workspace-id>.eu-west-1.staging-xata.dev/db/dbname:404_branch'
```

```bash
{"id":"60d017f9-b8ba-9211-974e-a0963da80ecd","message":"database branch [localxb:404_branch] not found"}%
```

CLI prints the error

cli/src/commands/migration/rollback.ts
divyenduz: Done, thanks for the feedback!

cli/src/commands/migration/start.ts
divyenduz: This is already a util right? 🤔 Or do you mean copying/ sharing the logic till this function call? I would personally prefer it like this (easier to read / follow / debug IMO).

cli/src/commands/migration/status.test.ts
divyenduz: Good idea, on it

SferaDev:
Added this PR as a topic for discussion in next week's pgroll sync

cli/src/commands/migration/start.ts
eemmiillyy: Sorry I commented on the wrong line, I meant `isActiveMigration`. Carry on :D

eemmiillyy:
I think the commands are run with `xata migration [...]` since they're nested under the `/migration` folder, just FYI. Not sure if we actually want them under a `/migrate` folder so the command is `xata migrate [...]`

eemmiillyy (approved):

.changeset/thick-bags-grab.md
```diff
@@ -1,5 +1,5 @@
 ---
-"@xata.io/cli": patch
+'@xata.io/cli': patch
```
divyenduz, line 2: ha, I think prettier did that when I was looking at where changesets are. But should be fine

andrew-farries (approved):

cli/src/commands/migration/start.ts
divyenduz: Ah it was this that was causing the build CI to fail with

`Invalid count value`

.changeset/thick-bags-grab.md
divyenduz: Ah, tracking PR that's why https://github.com/xataio/client-ts/pull/1547

PR #1543, opened 2024-07-23
feat: setup autoupdates for homebrew formula

eemmiillyy:


.github/workflows/release-cli-assets.yml
```diff
@@ -166,3 +201,83 @@ jobs:
...
+            cat ./Formula/xata.rb
+            echo
+            echo EOF
+          } >> "$GITHUB_OUTPUT"
+
+      - name: commit changes
+        run: |
+          status="Update dependencies to version ${{ needs.release-cli-assets.outputs.version }}"
+          git status
+          git commit -a -m "$status"
```
SferaDev, line 275: We can probably inline

.github/workflows/release-cli-assets.yml
```diff
@@ -166,3 +201,83 @@ jobs:
...
+
+      - name: Push changes
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GIT_TOKEN }}
+          branch: main
+          repository: xataio/homebrew-brew
+          force: true
```
SferaDev, line 283: Not sure if we should force push in this case, I'd prefer the action to fail and manually fix since this will be what a user downloads

SferaDev (approved):

PR #1572, opened 2024-09-12
Move env variable

SferaDev:
Move signing password to the correct build step


.changeset/honest-goats-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+@xata.io/cli: patch
```
SferaDev, line 2:
```suggestion
"@xata.io/cli": patch
```

divyenduz (approved):

PR #1573, opened 2024-09-12
Fix/csv import existing table

eemmiillyy:
Related: https://xata-hq.slack.com/archives/C032EKCBUGM/p1726051389626059

This PR adds the ability for Postgres users to import new data to existing tables. Before this PR csv import would error if the table already existed.

### DIFF
--- a/cli/src/utils/compareSchema.ts
+++ b/cli/src/utils/compareSchema.ts
@@ -59,45 +59,6 @@ export function compareSchemas({
     for (const column of deletedColumns) {
       edits.push({ drop_column: { table, column } });
     }
-
-    // Compare column properties
-    // for (const column of targetColumns) {
-    //   const sourceProps = source.tables?.[table]?.columns?.[column] ?? {};
-    //   const targetProps = target.tables?.[table]?.columns?.[column] ?? {};
-
-    //   if (sourceProps.type !== targetProps.type) {
-    //     edits.push({
-    //       alter_column: {
-    //         table,
-    //         column,
-    //         type: targetProps.type,
-    //         references:
-    //           targetProps?.type === 'link' && targetProps?.name
-    //             ? generateLinkReference({
-    //                 column: targetProps.name,
-    //                 table: tableNameFromLinkComment(targetProps?.comment ?? '') ?? ''
-    //               })
-    //             : undefined
-    //       }
-    //     });
-    //   }
-
-    //   if (sourceProps.nullable !== targetProps.nullable) {
-    //     edits.push({ alter_column: { table, column, nullable: targetProps.nullable } });
-    //   }
-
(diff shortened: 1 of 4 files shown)


cli/src/utils/compareSchema.ts
```diff
@@ -53,43 +61,43 @@ export function compareSchemas(
...
-              targetProps?.type === 'link' && targetProps?.name
-                ? generateLinkReference({
-                    column: targetProps.name,
-                    table: tableNameFromLinkComment(targetProps?.comment ?? '') ?? ''
-                  })
-                : undefined
-          }
-        });
-      }
+    // for (const column of targetColumns) {
```
eemmiillyy, line 64: I think we can maybe remove this bit. It's causing issues and we don't make use of it AFAICT

cli/src/utils/compareSchema.ts
```diff
@@ -127,3 +135,92 @@ export function compareSchemas(
 
   return { edits };
 }
+
+export const inferOldSchemaToNew = (oldSchema: Schemas.DBBranch): Schemas.BranchSchema => {
```
eemmiillyy, line 139: This will be shared between frontend when https://github.com/xataio/frontend-next/issues/3873 is taken care of

cli/src/utils/compareSchema.ts
```diff
@@ -51,45 +59,6 @@ export function compareSchemas(
     for (const column of deletedColumns) {
       edits.push({ drop_column: { table, column } });
     }
-
-    // Compare column properties
-    for (const column of targetColumns) {
```
eemmiillyy, line 56: In order to get this working properly we would need to:;
1. add the correct up/down fields for the migrations
2. split apart add columns and alter columns into separate migrations to run because of the limitation of altering new objects from pgroll

I removed it for now

cli/src/utils/compareSchema.ts
```diff
@@ -17,8 +21,12 @@ export function compareSchemas(
 
   // Compare columns
   for (const table of sourceTables) {
-    const sourceColumns = Object.keys(source.tables?.[table]?.columns ?? {});
-    const targetColumns = Object.keys(target.tables?.[table]?.columns ?? {});
+    const sourceColumns = Object.keys(source.tables?.[table]?.columns ?? {}).filter(
```
eemmiillyy, line 24: Since we use this function with `adaptTable` = true, adapt will add the internal columns to the migration if they don't exist causing it to fail.

divyenduz (approved):
LGTM, CI is failing though.

eemmiillyy:
I think the failures are unrelated, asked about it here: https://xata-hq.slack.com/archives/C05UZ09G23S/p1726148857070869 @divyenduz

PR #1589, opened 2024-09-19
Remove ifVersion

SferaDev:
Remove `ifVersion` feature as part of the Data API deprecation process


eemmiillyy (approved):
Is this for kysely repo only or existing repo aswell?

SferaDev:
For both @eemmiillyy

